PR #115, opened 2026-01-06
feat(csharp): align project structures and output locations to other drivers

davidhcoe:
## What's Changed

- aligns the project structure and output location (artifacts) with the other ADBC drivers
- added Directory.Build.props file


eric-wang-1990:
It seems this is from your forked PR which cannot get the env variables, so all the tests failed, but looks good otherwise.

eric-wang-1990 (approved):

PR #155, opened 2026-01-20
Spec for metadata refactor in ADBC to use a common source of truth for SEA and thrift and to use the same schema builder logic

msrathore-db:
## What's Changed
This is the plan for refactoring the metadata methods in upstream arrow repo and the databricks driver repo. The refactoring should not affect the behaviour for thrift for the other drivers.

### DIFF
--- a/metadata_refactor.md
+++ b/metadata_refactor.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
-Refactor ADBC C# metadata implementation to eliminate ~60% code duplication between HiveServer2 (Thrift) and StatementExecution API (SEA) by creating shared abstractions for type mapping, schema construction, and field synthesis across **ALL metadata calls** (GetColumns, GetPrimaryKeys, GetCrossReference, etc.).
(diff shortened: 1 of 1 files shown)


metadata_refactor.md
```diff
@@ -0,0 +1,632 @@
...
+
+**Methods**:
+
+**For Simple Metadata** (used by both GetObjects hierarchical and statement-based flat):
+- `BuildFlatCatalogsSchema(IEnumerable<CatalogMetadataRecord>)` - 1-column JDBC-style schema (TABLE_CAT)
+- `BuildFlatSchemasSchema(IEnumerable<SchemaMetadataRecord>)` - 2-column JDBC-style schema (TABLE_CATALOG, TABLE_SCHEM)
+- `BuildFlatTablesSchema(IEnumerable<TableMetadataRecord>)` - 10-column JDBC-style schema (TABLE_CAT, TABLE_SCHEM, TABLE_NAME, TABLE_TYPE, REMARKS, ...)
+
+**For Column Metadata**:
+- `BuildHierarchicalColumnsSchema(IEnumerable<ColumnMetadataRecord>)` - 19-field nested structure for GetObjects
```
eric-wang-1990: This is kind of misleading, the 19-field is the leaf fields right? What it returns is actually a hierachy which starts from the catalog level?

PR #183, opened 2026-01-23
feat(csharp): add straggler download detection and mitigation for CloudFetch

msrathore-db:
## Summary

Adds the straggler download detection and mitigation feature for CloudFetch, along with comprehensive unit and E2E test coverage. This feature identifies slow ("straggler") downloads in parallel CloudFetch batches, cancels and retries them, and optionally falls back to sequential downloading when too many stragglers are detected.

### New Files

| File | Description |
|---|---|
| `CloudFetchStragglerMitigationConfig.cs` | Configuration class with `Parse()` factory that reads connection properties. Includes validation and a static `Disabled` singleton. |
| `FileDownloadMetrics.cs` | Per-file timing/throughput tracker. Thread-safe via lock. Tracks start time, end time, completion, and straggler-cancelled state. |
| `StragglerDownloadDetector.cs` | Core detection engine. Runs a background monitoring loop every 2s, uses median throughput from completed downloads to identify stragglers, cancels them via per-file `CancellationTokenSource`, and manages sequential fallback. |

### Modified Files

| File | Change |
|---|---|
| `DatabricksParameters.cs` | Added 6 new connection parameters (see Configuration section below) |
| `CloudFetchConfiguration.cs` | Added `StragglerMitigationConfig` property, parsed from connection properties |
| `CloudFetchDownloader.cs` | Integrated straggler detection into the download pipeline — registers downloads, handles straggler cancellation with retry + exponential backoff, acquires sequential permits when in fallback mode, and ensures resource cleanup on failure |

## Configuration Parameters

| Parameter | Default | Description |
|---|---|---|
| `adbc.databricks.cloudfetch.straggler_mitigation_enabled` | `false` | Master toggle for straggler detection |
| `adbc.databricks.cloudfetch.straggler_multiplier` | `1.5` | Download is a straggler if elapsed > multiplier × expected time |
| `adbc.databricks.cloudfetch.straggler_quantile` | `0.6` | Detection activates only after this fraction of downloads complete |
| `adbc.databricks.cloudfetch.straggler_padding_seconds` | `5` | Buffer time (seconds) added before declaring a straggler |
| `adbc.databricks.cloudfetch.max_stragglers_per_query` | `10` | Max stragglers before triggering sequential fallback |
| `adbc.databricks.cloudfetch.synchronous_fallback_enabled` | `false` | Whether to switch to sequential downloads after threshold |

## Detection Algorithm

1. A background loop runs every **2 seconds** while downloads are active
2. Detection only activates after the **quantile threshold** (default 60%) of downloads in the batch have completed — prevents premature detection with insufficient data
3. Calculates **median throughput** (bytes/sec) from completed downloads — robust to outliers compared to mean
4. For each active download, computes expected time: `(multiplier × fileSize / medianThroughput) + padding`
5. If `elapsed > expectedTime`, the download is marked as a straggler and its per-file `CancellationToken` is cancelled
6. Cancelled stragglers are **retried with exponential backoff** (doubling up to 32s cap, with 80-120% jitter)
7. If the straggler count exceeds `max_stragglers_per_query` and `synchronous_fallback_enabled` is true, all subsequent downloads in the batch are gated to **one-at-a-time** via a `SemaphoreSlim(1,1)` permit

## Design Decisions

- **Feature is off by default** — both `straggler_mitigation_enabled` and `synchronous_fallback_enabled` must be explicitly enabled
- **Per-file CancellationToken pattern** — each download gets a linked CTS so the monitor can cancel individual downloads without affecting others or the global token
- **Per-query scope** — straggler state and sequential fallback are scoped to a single downloader instance; a new batch starts fresh in parallel mode
- **RAII-style `SequentialDownloadPermit`** — uses `Interlocked.CompareExchange` for idempotent disposal, with a static `NoOp` singleton for zero overhead when not in fallback mode
- **Duplicate detection prevention** — `_alreadyCountedStragglers` dictionary persists for the entire query to prevent re-counting a straggler after retry
- **Race-safe completion** — if a download completes just as the monitor cancels it, the already-obtained data is used successfully

## Thread Safety

- `FileDownloadMetrics`: private lock object for all mutable state
- `StragglerDownloadDetector`: `ConcurrentDictionary` for all collections, `Interlocked` for atomic counter, `volatile` for fallback flag
- Monitoring loop handles `ObjectDisposedException` from CTS race between cancellation and download completion
- Resource leak prevention: try/catch around task creation ensures sequential permit, download semaphore, and memory are released on failure

### Unit Tests — `CloudFetchStragglerUnitTests.cs` (622 lines, 19 tests)

| Category | Tests |
|---|---|
| Configuration | Default values, custom values, validation, builder |
| FileDownloadMetrics | Throughput before/after completion, straggler flag |
| Detection Algorithm | Median calculation (odd/even), no completed downloads, below quantile threshold, fallback threshold |
| Duplicate Prevention | Same file counted once across multiple detection cycles |
| CTS Management | Atomic replacement, concurrent cleanup |
| Cleanup Behavior | Executes in finally block, respects shutdown cancellation |
| Edge Cases | Counter overflow (long), concurrent modification safety |

### E2E Tests — `CloudFetchStragglerDownloaderE2ETests.cs` (1491 lines, 19 tests)

Integration tests using mocked HTTP handlers with the actual `CloudFetchDownloader`:

| Category | Tests |
|---|---|
| Configuration | Feature disabled by default, parameter names and definitions |
| Detection & Retry | Slow download identified and cancelled, fast downloads not flagged, minimum quantile required, cancelled straggler retried, mixed speed downloads |
| Sequential Fallback | Activates after threshold, enforces one-at-a-time, no false positives in sequential mode, fallback scoped to current batch |
| Behavior | Monitoring respects cancellation, parallel mode respects max downloads, clean shutdown |
| Component Creation | Detector and metrics objects initialize correctly, counter increments atomically |

## Manual Testing

Manual testing was performed with code modified to introduce stragglers — successfully reproduced both straggler detection and sequential fallback on Power BI.


csharp/src/Reader/CloudFetch/CloudFetchDownloader.cs
```diff
@@ -552,9 +601,48 @@ await _activityTracer.TraceActivityAsync(async activity =>
                         fileData = await response.Content.ReadAsByteArrayAsync().ConfigureAwait(false);
```
eric-wang-1990: ⏺ This is worth clarifying — HttpCompletionOption.ResponseHeadersRead is specifically designed to not download the body during SendAsync. That's the whole purpose of the flag.                                                                  
                                                                                                                                                                                                                                                 
  The timeline is:                                                                                                                                                                                                                               
  SendAsync(..., ResponseHeadersRead, effectiveToken)                                                                                                                                                                                            
    → sends HTTP GET request                                                                                                                                                                                                                     
    → waits for response HEADERS only                                                                                                                                                                                                            
    → returns HttpResponseMessage as soon as headers arrive
    → body data has NOT been received yet

  response.Content.ReadAsByteArrayAsync()
    → THIS is where the TCP socket reads happen
    → bytes of the file come over the network here
    → for a 100MB CloudFetch file, all 100MB transfer happens here

  ResponseHeadersRead is specifically chosen here because it gives the code a chance to inspect the response status code (check for 403/expired URL) before committing to reading the potentially large body. It's a deliberate split: fast
  header check first, then body read.

  The straggler would be detected during ReadAsByteArrayAsync — that's when a slow connection would manifest. The SendAsync with ResponseHeadersRead returns relatively quickly (just TCP + HTTP headers), so the monitoring thread would almost
  never fire during that window.

  So the fix of passing effectiveToken to ReadAsByteArrayAsync(effectiveToken) is still the right one — that's where the slow work is actually happening and where the cancellation needs to land.

csharp/src/Reader/CloudFetch/CloudFetchDownloader.cs
```diff
@@ -398,8 +434,17 @@ await _activityTracer.TraceActivityAsync(async activity =>
                                 }
                             }, cancellationToken);
```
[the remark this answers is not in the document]
eric-wang-1990: Bump this

csharp/src/Reader/CloudFetch/CloudFetchDownloader.cs
```diff
@@ -621,6 +675,52 @@ await _activityTracer.TraceActivityAsync(async activity =>
...
+                        // Check if we would exceed the time budget
+                        if (retryTimeoutMs > 0 && totalRetryWaitMs + waitMs > retryTimeoutMs)
+                        {
+                            activity?.AddEvent("cloudfetch.straggler_retry_timeout_exceeded", [
+                                new("offset", downloadResult.StartRowOffset),
+                                new("sanitized_url", sanitizedUrl),
+                                new("total_attempts", attemptCount),
+                                new("total_retry_wait_ms", totalRetryWaitMs)
+                            ]);
+                            break;
```
eric-wang-1990, line 726: `lastException` is `null` when this `break` fires, so the error thrown after the loop will say `Last error: unknown` instead of naming the actual cause.

The normal retry timeout path (below, in `catch (Exception ex)`) sets `lastException` before breaking, so the real exception type appears in the message. The straggler path skips that `catch` entirely.

Fix: set `lastException` before breaking:
```csharp
lastException = new InvalidOperationException(
    $"Straggler retry budget exhausted at offset {downloadResult.StartRowOffset} after {attemptCount} cancellation(s).");
break;
```

csharp/src/Reader/CloudFetch/StragglerDownload/StragglerDownloadDetector.cs
```diff
@@ -0,0 +1,640 @@
...
+
+            // Cleanup remaining per-file cancellation tokens
+            foreach (var cts in _perFileDownloadCancellationTokens.Values)
+            {
+                cts?.Dispose();
+            }
+            _perFileDownloadCancellationTokens.Clear();
+
+            // Clear completed download metrics history
+            _completedDownloadMetrics.Clear();
```
eric-wang-1990, line 188: `_sequentialSemaphore` is not disposed here, leaking one `SemaphoreSlim` per query when straggler mitigation is enabled.

`StopMonitoring()` cleans up `_monitoringCts` and the per-file CTSes but skips the semaphore. `_sequentialSemaphore` is only disposed in `Dispose()`, but `Dispose()` is never called — `CloudFetchDownloadManager` holds `_downloader` as `ICloudFetchDownloader` which does not extend `IDisposable`.

Minimal fix for the happy path: add `_sequentialSemaphore?.Dispose();` here in `StopMonitoring()`.

Full fix: also add `IDisposable` to `ICloudFetchDownloader`, implement it on `CloudFetchDownloader`, and call `_downloader.Dispose()` in `CloudFetchDownloadManager.Dispose()` to cover exception paths that bypass `StopAsync`.

csharp/src/Reader/CloudFetch/CloudFetchDownloader.cs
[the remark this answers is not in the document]
msrathore-db: Addressed both items:

**1(b) — `cancellationToken` in `ContinueWith`:** Good catch — this is a real bug. When the token is cancelled, the continuation body is skipped entirely, so `_downloadSemaphore.Release()` and `sequentialPermit?.Dispose()` never run. Removed `cancellationToken` from `ContinueWith` so the continuation always executes.

**1(a) — try/catch around task creation:** I'd push back on moving this into `ContinueWith`. The outer try/catch is a defensive safety net for the (admittedly rare) case where task creation itself fails. Note the catch already uses `SetError()` + `break`, not `throw` — it follows the existing pattern. Moving the resource cleanup into `ContinueWith` wouldn't help for this case, and the memory release (`_memoryManager.ReleaseMemory`) is intentionally handled inside `DownloadFileAsync` after all retries, not in the continuation — moving it to `ContinueWith` would risk double-releasing.

csharp/src/Reader/CloudFetch/CloudFetchDownloader.cs
msrathore-db: Thanks for the detailed explanation — great context on the `ResponseHeadersRead` split.

The code already handles this correctly: `effectiveToken` is propagated to the body read via `bodyTimeoutCts` (a linked token created from `effectiveToken` when straggler detection is enabled — see `CancellationToken baseToken = _stragglerDetector != null ? effectiveToken : cancellationToken`). The body is read with `CopyToAsync(memoryStream, 81920, bodyTimeoutCts.Token)` rather than `ReadAsByteArrayAsync`, specifically because the latter lacks a `CancellationToken` overload on net472. So straggler cancellation does land on the body transfer where slow connections actually manifest.

csharp/src/Reader/CloudFetch/CloudFetchDownloader.cs
msrathore-db: Good catch. Fixed — added `lastException` assignment before the break:

```csharp
lastException = new OperationCanceledException(
    $"Straggler retry budget exhausted at offset {downloadResult.StartRowOffset} after {attemptCount} cancellation(s).");
```

This only manifests when all retry failures are straggler cancellations (if any regular exception occurs during retries, `lastException` gets set in the generic catch block).

csharp/src/Reader/CloudFetch/StragglerDownload/StragglerDownloadDetector.cs
msrathore-db: Good catch. Added `_sequentialSemaphore.Dispose()` to `StopMonitoring()` alongside the other cleanup (per-file CTSes and completed metrics).

For the broader `IDisposable` chain (`ICloudFetchDownloader` → `CloudFetchDownloader` → `CloudFetchDownloadManager.Dispose()`): `SemaphoreSlim` without `AvailableWaitHandle` accessed only holds managed memory (no kernel handles), so the practical impact here was negligible — but agreed it's proper hygiene. I'll address the full `IDisposable` wiring in a follow-up to keep this PR scoped.

msrathore-db: Updated the fix — the initial approach of disposing `_sequentialSemaphore` in `StopMonitoring()` was wrong because `StopAsync()` calls `StopMonitoring()` *before* awaiting `_downloadTask`. In-flight downloads may still have `ContinueWith` callbacks pending that call `sequentialPermit.Dispose()` → `_semaphore.Release()`, which would throw `ObjectDisposedException` on the already-disposed semaphore.

Corrected approach: call `_stragglerDetector.Dispose()` in the `finally` block of `StopAsync()`, **after** `_downloadTask` completes — so all permits have safely released the semaphore first.

Lifecycle is now:
1. `StopMonitoring()` — stops the background monitoring loop, cleans up CTSes
2. `await _downloadTask` — all in-flight downloads finish, permits release the semaphore
3. `_stragglerDetector.Dispose()` — safely disposes the semaphore (in `finally`)

csharp/src/Reader/CloudFetch/CloudFetchDownloader.cs
msrathore-db: Pushed the fix in `a3c0295`.

csharp/src/Reader/CloudFetch/CloudFetchDownloader.cs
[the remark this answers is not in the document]
msrathore-db: Both fixes pushed in `a3c0295`.

csharp/src/Reader/CloudFetch/CloudFetchDownloader.cs
msrathore-db: All fixes pushed in `a3c0295`.

csharp/src/Reader/CloudFetch/CloudFetchDownloader.cs
```diff
@@ -398,8 +434,17 @@ await _activityTracer.TraceActivityAsync(async activity =>
...
+                            // Add the task to the dictionary
+                            downloadTasks[downloadTask] = downloadResult;
+                        }
+                        catch
+                        {
+                            // If task creation fails, release all acquired resources to prevent leak
+                            sequentialPermit?.Dispose();
+                            _downloadSemaphore.Release();
+                            _memoryManager.ReleaseMemory(size);
+                            throw;
```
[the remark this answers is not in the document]
msrathore-db: I'd push back on moving this into `ContinueWith`. The outer try/catch is a defensive safety net for the edge case where task creation itself fails (e.g., OOM from `ContinueWith`). In practice it's dead code — `DownloadFileAsync` is `async Task`, so exceptions are captured into the returned Task, never thrown synchronously.

But moving the cleanup into `ContinueWith` would be wrong: `_memoryManager.ReleaseMemory(size)` is intentionally handled inside `DownloadFileAsync` itself (after all retries exhaust), not in the continuation. If we added it to `ContinueWith`, we'd get a double release when the download task faults.

Also note the catch already uses `SetError()` + `break` (not `throw`) — it does follow the existing pattern.

Separately, the `cancellationToken` on `ContinueWith` was a real bug (see reply above) — that's fixed.

csharp/src/DatabricksParameters.cs
```diff
@@ -182,6 +182,50 @@ public class DatabricksParameters : SparkParameters
...
+        /// When enabled, the driver monitors download speeds and cancels stragglers to retry them.
+        /// Default value is false if not specified.
+        /// </summary>
+        public const string CloudFetchStragglerMitigationEnabled = "adbc.databricks.cloudfetch.straggler_mitigation_enabled";
+
+        /// <summary>
+        /// Straggler throughput multiplier for CloudFetch operations.
+        /// A download is considered a straggler if it takes more than (multiplier × expected_time) to complete.
+        /// Default value is 1.5 if not specified.
+        /// Must be greater than or equal to 1.0.
```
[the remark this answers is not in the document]
msrathore-db: Fixed in d98b722 — XML doc now reads "Must be greater than 1.0" to match the runtime check at `StragglerDownloadDetector.cs:68` (`config.Multiplier <= 1.0` rejects). Multiplier=1.0 would mean "a download is a straggler iff slower than median," which isn't a useful signal, so the validation is correct and the doc was wrong.

csharp/src/Reader/CloudFetch/StragglerDownload/StragglerDownloadDetector.cs
```diff
@@ -0,0 +1,640 @@
...
+                            var download = allDownloadMetrics.FirstOrDefault(m => m.FileOffset == offset);
+                            if (download != null)
+                            {
+                                TimeSpan elapsed = currentTime - download.DownloadStartTime;
+                                double elapsedSeconds = elapsed.TotalSeconds;
+
+                                activity.AddEvent("cloudfetch.straggler_identified", [
+                                    new("offset", download.FileOffset),
+                                    new("elapsed_seconds", elapsedSeconds),
+                                    new("file_size_mb", download.FileSizeBytes / (1024.0 / 1024.0))
```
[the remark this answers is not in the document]
msrathore-db: Fixed in d98b722 — `(1024.0 / 1024.0)` was indeed dividing by 1, so the event was logging bytes labeled as MB. Changed to `(1024.0 * 1024.0)`. Verified other `/ 1024.0 / 1024.0` usages in the file (`CloudFetchDownloader.cs:497, 702`) are correct (left-to-right division).

csharp/src/Reader/CloudFetch/StragglerDownload/StragglerDownloadDetector.cs
```diff
@@ -0,0 +1,640 @@
...
+            _totalStragglersDetectedInQuery = 0;
+            _hasTriggeredSequentialDownloadFallback = false;
+        }
+
+        /// <summary>
+        /// Gets a value indicating whether the query should fall back to sequential downloads
+        /// due to exceeding the maximum straggler threshold.
+        /// </summary>
+        public bool ShouldFallbackToSequentialDownloads =>
+            _totalStragglersDetectedInQuery >= (_config.SynchronousFallbackEnabled ? _config.MaxStragglersBeforeFallback : long.MaxValue);
```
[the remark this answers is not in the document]
msrathore-db: Fixed in d98b722 — `ShouldFallbackToSequentialDownloads` now uses `Interlocked.Read(ref _totalStragglersDetectedInQuery)`, consistent with `GetTotalStragglersDetectedInQuery()` at line 453. Worth noting the practical impact is small even on 32-bit (worst case is one extra parallel download slipping through one 2s monitor cycle, then self-correcting), but the consistency is the right reason to fix it.

csharp/test/E2E/CloudFetch/CloudFetchStragglerDownloaderE2ETests.cs
```diff
@@ -0,0 +1,1491 @@
...
+                    try
+                    {
+                        await Task.Delay(delayMs, token);
+                    }
+                    finally
+                    {
+                        if (offset > 0)
+                        {
+                            concurrentDownloads.TryRemove(offset, out _);
+                        }
```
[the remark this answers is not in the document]
msrathore-db: Fixed in d98b722 — dropped the `if (offset > 0)` guard. `file0` is a valid first chunk and was being added unconditionally but never removed, leaving a permanent stale entry that biased `maxConcurrency` assertions. `TryRemove` is already idempotent so the guard added nothing.

csharp/test/E2E/CloudFetch/CloudFetchStragglerDownloaderE2ETests.cs
```diff
@@ -0,0 +1,1491 @@
...
+                        downloadCancelledFlags[offset] = true;
+                        throw;
+                    }
+                    finally
+                    {
+                        if (offset > 0)
+                        {
+                            concurrentDownloads.TryRemove(offset, out _);
+                        }
+                    }
```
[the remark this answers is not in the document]
msrathore-db: Fixed in d98b722 — same fix applied here in the second helper (dropped the `if (offset > 0)` guard). Acknowledging the copy-paste between the two helpers is a smell; will consolidate them in a follow-up to keep this PR focused.

csharp/src/Reader/CloudFetch/CloudFetchDownloader.cs
```diff
@@ -621,6 +682,54 @@ await _activityTracer.TraceActivityAsync(async activity =>
...
+                        activity?.AddEvent("cloudfetch.straggler_cancelled", [
+                            new("offset", downloadResult.StartRowOffset),
+                            new("sanitized_url", sanitizedUrl),
+                            new("file_size_mb", size / 1024.0 / 1024.0),
+                            new("elapsed_seconds", stopwatch.ElapsedMilliseconds / 1000.0),
+                            new("attempt", attemptCount),
+                            new("max_retries", _maxRetries)
+                        ]);
+
+                        _stragglerDetector?.MarkCancelledAsStragler(fileOffset, activity);
```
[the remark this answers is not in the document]
msrathore-db: Deferring this rename to a follow-up PR. Confirmed at `StragglerDownloadDetector.cs:266` that the externally-visible telemetry attribute is already spelled correctly (`was_cancelled_as_straggler`), so no log/metrics consumer is impacted by the internal misspelling. `MarkCancelledAsStragler`, `WasCancelledAsStragler` are all internal; pure rename, but the diff churn isn't worth it on this PR which is already large.

csharp/src/Reader/CloudFetch/StragglerDownload/FileDownloadMetrics.cs
```diff
@@ -0,0 +1,137 @@
...
+        /// <summary>
+        /// Marks this download as having been cancelled due to being identified as a straggler.
+        /// Thread-safe - idempotent.
+        /// </summary>
+        public void MarkCancelledAsStragler()
+        {
+            lock (_lock)
+            {
+                _wasCancelledAsStragler = true;
+            }
```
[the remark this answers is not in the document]
msrathore-db: Deferring to a follow-up PR — same reasoning as the call-site reply: the external telemetry attribute (`was_cancelled_as_straggler` at `StragglerDownloadDetector.cs:266`) is already spelled correctly, and `MarkCancelledAsStragler`/`WasCancelledAsStragler` are internal-only. Pure rename, will batch with other internal cleanups.

csharp/src/Reader/CloudFetch/CloudFetchDownloader.cs
```diff
@@ -361,16 +392,29 @@ await _activityTracer.TraceActivityAsync(async activity =>
...
+                        Task downloadTask;
+                        try
+                        {
+                            // Start the download task
+                            downloadTask = DownloadFileAsync(downloadResult, cancellationToken)
+                                .ContinueWith(t =>
+                                {
+                                    // Release in reverse order (LIFO)
+                                    sequentialPermit?.Dispose();
+                                    _downloadSemaphore.Release();
```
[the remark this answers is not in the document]
msrathore-db: Pushing back on this one. The download loop runs inside a `Task.Run` from `CloudFetchDownloadManager` — there is no captured `SynchronizationContext` in scope when this `ContinueWith` is registered, so the continuation will run on the thread pool by default. Adding `TaskScheduler.Default` here would be defensive but redundant; auditing the rest of the driver for the same general best-practice would be a separate, broader change. Happy to add it as a one-liner if you'd prefer the explicit hint, but I don't think it's blocking.

csharp/test/E2E/CloudFetch/CloudFetchStragglerDownloaderE2ETests.cs
```diff
@@ -0,0 +1,1491 @@
...
+                catch (OperationCanceledException)
+                {
+                    // Expected during cleanup
+                }
+            });
+
+            // Wait for monitoring to detect straggler (need 6/10 completions, then detection)
+            // Monitoring runs every 2 seconds, so wait at least 4-5 seconds to allow multiple checks
+            await Task.Delay(5000);
+
```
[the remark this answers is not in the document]
msrathore-db: Deferring the time-source injection refactor to a follow-up. Two reasons: (1) the detector's correctness IS timing-based — its job is "did >Nms elapse since download start" — so a fake-clock test would mostly verify "did we call IClock correctly" rather than "does the monitor loop behave under real timing"; (2) injecting a time source means changing the public ctor of `StragglerDownloadDetector` and threading it through `CloudFetchDownloader`, which is a non-trivial API change.

Will tag these tests with a Slow trait and ticket the refactor so we can decide on the right approach (likely also extracting `StragglerMonitoringInterval` to be configurable) without blocking this PR.

msrathore-db:
### Round of review fixes pushed in d98b722

**Fixed:**
- Multiplier XML doc now matches validation (`> 1.0` not `>= 1.0`)
- `StragglerDownloadDetector.ShouldFallbackToSequentialDownloads` uses `Interlocked.Read` for consistency with `GetTotalStragglersDetectedInQuery()`
- `file_size_mb` math in `straggler_identified` event (was logging bytes labeled as MB)
- E2E concurrency-helper bug: `offset > 0` guard left a stale `file0` entry, biasing `maxConcurrency` assertions in two helpers
- **`downloadTasks` re-keyed by `ChunkIndex`** — Copilot's suppressed comment was correct: the dict was keyed by the continuation Task but the `TryRemove(t, ...)` call inside the continuation receives the antecedent task as `t`, so removals never landed and the dict accumulated all tasks for the lifetime of the download loop. Now keyed by `downloadResult.ChunkIndex` (`ConcurrentDictionary<long, Task>`); `Task.WhenAll` uses `.Values`.

**Deferred to follow-up PR (replies on individual threads):**
- `Stragler` → `Straggler` internal rename. Confirmed at `StragglerDownloadDetector.cs:266` that the shipping telemetry attribute (`was_cancelled_as_straggler`) is already spelled correctly, so no external consumer is affected.
- `new Random()` in jitter (`CloudFetchDownloader.cs:713, 738`). `Random.Shared` isn't on net472, so the right fix is `[ThreadStatic] static Random`. Same pattern also at `RetryHttpHandler.cs:313`. Practical impact is small (backoff doubling diverges retries after one step).
- E2E tests using real-time `Task.Delay`. Will tag as Slow and ticket the time-source-injection refactor (also requires making `StragglerMonitoringInterval` configurable).

**Pushed back on:**
- Adding `TaskScheduler.Default` to the `ContinueWith`. The download loop runs inside `Task.Run` with no captured sync context, so the continuation already runs on the thread pool. Happy to add as defense-in-depth if reviewers want it.

**Verification:** driver + tests build clean (0 errors / 0 warnings). All 38 `CloudFetchStraggler` tests pass.

csharp/src/Reader/CloudFetch/CloudFetchDownloader.cs
```diff
@@ -286,8 +317,8 @@ await _activityTracer.TraceActivityAsync(async activity =>
 
                 try
                 {
-                    // Keep track of active download tasks
-                    var downloadTasks = new ConcurrentDictionary<Task, IDownloadResult>();
+                    // Keep track of active download tasks keyed by ChunkIndex
+                    var downloadTasks = new ConcurrentDictionary<long, Task>();
```
eric-wang-1990: Is the IDownloadResult used in any ways currently?

csharp/src/Reader/CloudFetch/CloudFetchDownloader.cs
```diff
@@ -630,6 +692,54 @@ await _activityTracer.TraceActivityAsync(async activity =>
...
+                            new("attempt", attemptCount),
+                            new("max_retries", _maxRetries)
+                        ]);
+
+                        _stragglerDetector?.MarkCancelledAsStragler(fileOffset, activity);
+                        effectiveToken = _stragglerDetector?.RegisterDownload(fileOffset, size, cancellationToken, activity)
+                                       ?? cancellationToken;
+
+                        // Exponential backoff with jitter (80-120% of base)
+                        int waitMs = (int)Math.Max(100, currentBackoffMs * (0.8 + new Random().NextDouble() * 0.4));
```
eric-wang-1990, line 722: **Bug:** `new Random()` created on every retry iteration produces correlated values in older .NET when called in rapid succession. Use `Random.Shared.NextDouble()` instead. Same issue at line 748.

csharp/src/Reader/CloudFetch/StragglerDownload/StragglerDownloadDetector.cs
```diff
@@ -0,0 +1,640 @@
...
+            // Filter out downloads that were already detected as stragglers (prevents re-detection on retry)
+            var activeDownloads = allDownloadMetrics.Where(m => !m.IsDownloadCompleted && !countingDict.ContainsKey(m.FileOffset)).ToList();
+
+            if (activeDownloads.Count == 0)
+            {
+                return Enumerable.Empty<long>();
+            }
+
+            // Check if we have enough completed downloads to calculate median
+            int totalDownloads = allDownloadMetrics.Count + _completedDownloadMetrics.Count;
```
eric-wang-1990: **Bug:** `totalDownloads` double-counts downloads that have just completed — they appear in both `allDownloadMetrics` (as completed entries) and `_completedDownloadMetrics`. This inflates `requiredCompletions` and delays straggler detection.

csharp/src/Reader/CloudFetch/StragglerDownload/StragglerDownloadDetector.cs
```diff
@@ -0,0 +1,640 @@
...
+
+            // Immediately remove and dispose CTS
+            // Race condition with monitoring thread is handled by try-catch in MonitoringLoopAsync
+            if (_perFileDownloadCancellationTokens.TryRemove(fileOffset, out var cts))
+            {
+                cts?.Dispose();
+            }
+        }
+
+        public void Dispose()
```
eric-wang-1990, line 612: **Bug:** `Dispose()` cancels `_monitoringCts` but doesn't await `_monitoringTask`. If the monitoring loop is mid-execution (e.g. calling `cts.Cancel()` on per-file tokens), it may access state that's already been disposed. `StopMonitoring()` handles this correctly — either call it from `Dispose()`, or document that callers must invoke `StopMonitoring()` before `Dispose()`.

csharp/src/Reader/CloudFetch/StragglerDownload/FileDownloadMetrics.cs
```diff
@@ -0,0 +1,137 @@
...
+{
+    /// <summary>
+    /// Tracks timing and throughput metrics for individual file downloads.
+    /// Thread-safe for concurrent access.
+    /// </summary>
+    internal class FileDownloadMetrics
+    {
+        private readonly object _lock = new object();
+        private DateTime? _downloadEndTime;
+        private bool _wasCancelledAsStragler;
```
eric-wang-1990: Typo: "Stragler" → "Straggler" (missing second 'g'). Affects `_wasCancelledAsStragler`, `WasCancelledAsStragler`, and `MarkCancelledAsStragler` throughout this file and in `StragglerDownloadDetector.cs`.

csharp/src/Reader/CloudFetch/CloudFetchDownloader.cs
msrathore-db: Confirmed by reading every reference: the `IDownloadResult` value was never read out of `downloadTasks` in the old code. The five sites (`Count`, `Keys`, `TryRemove`, indexer-set, and the constructor) all either ignore the value or write it. Most notably, the continuation closure references `downloadResult` directly via foreach-variable capture (e.g., for `SetFailed(ex)`, the failed-download tracing, `totalBytes += downloadResult.Size`), not via lookup in the dict — so the dict's value field was redundant with the closure capture.

The new `<long, Task>` keeps the only thing actually needed (the Task, for `WhenAll(.Values)`) and uses `ChunkIndex` as the key so the continuation can correctly remove its own entry — which was the underlying bug fix here.

csharp/src/Reader/CloudFetch/CloudFetchDownloader.cs
msrathore-db: The `new Random()` here mirrors the existing jitter pattern already in this method's generic retry catch (line 748, introduced in #375 before this PR) — we copied that exact expression when adding the straggler catch block to keep the backoff semantics consistent across both retry paths. So this is matching the established convention in the file, not a fresh choice.

One caveat on the suggested fix: `Random.Shared` is .NET 6+ only — substituting it directly breaks the `net472` and `netstandard2.0` builds with CS0117. The portable fix is `[ThreadStatic] static Random _retryRng`, which works on all three target frameworks. Happy to apply that here (and also at line 748 since we'd be touching the same backoff formula) if you'd like the convention updated in this PR.

csharp/src/Reader/CloudFetch/StragglerDownload/FileDownloadMetrics.cs
msrathore-db: Fixing in this PR. Renaming `_wasCancelledAsStragler`, `WasCancelledAsStragler`, and `MarkCancelledAsStragler` → `Straggler` in `FileDownloadMetrics.cs` and `StragglerDownloadDetector.cs`, plus the call site at `CloudFetchDownloader.cs:718`. The external telemetry attribute at `StragglerDownloadDetector.cs:266` is already correctly spelled `was_cancelled_as_straggler`, so the rename is internal-only.

csharp/src/Reader/CloudFetch/StragglerDownload/StragglerDownloadDetector.cs
msrathore-db: Verified — race is real: a metric transitioning from active→completed between the line 394 snapshot and the line 337/349 reads gets counted in both `allDownloadMetrics` and `_completedDownloadMetrics`, inflating the totals.

Fixing by deduplicating on `FileOffset` — history entries already represented in the active snapshot are excluded from `completedFromHistory`, and `totalDownloads` is computed from the deduped count.

csharp/src/Reader/CloudFetch/StragglerDownload/StragglerDownloadDetector.cs
msrathore-db: Fixing — `Dispose()` now calls `StopMonitoring().GetAwaiter().GetResult()` first so the monitoring task is awaited before any resource teardown. `StopMonitoring` is idempotent (early-returns when `_monitoringTask` is null), so it's safe to call from both `StopAsync` and `Dispose`.

Verified no deadlock risk: in the current call chain, `Dispose` runs on a thread-pool thread (StopAsync awaits with `ConfigureAwait(false)` before reaching it), and the monitoring task itself never captures a SynchronizationContext (StartMonitoring is reached after a `ConfigureAwait(false)` upstream, and the loop's internal awaits use `ConfigureAwait(false)`).

csharp/src/Reader/CloudFetch/StragglerDownload/StragglerDownloadDetector.cs
eric-wang-1990: Is this fix pushed?

csharp/src/Reader/CloudFetch/StragglerDownload/StragglerDownloadDetector.cs
eric-wang-1990: Em I am not seeing the change?

csharp/src/Reader/CloudFetch/StragglerDownload/StragglerDownloadDetector.cs
```diff
@@ -0,0 +1,640 @@
...
+
+            // Add tracing only if activity is provided
+            if (activity != null)
+            {
+                // Calculate metrics for tracing only when needed
+                var completedDownloads = allDownloadMetrics.Where(m => m.IsDownloadCompleted).ToList();
+                var activeDownloads = allDownloadMetrics.Where(m => !m.IsDownloadCompleted && !m.WasCancelledAsStragler).ToList();
+                int requiredCompletions = (int)Math.Ceiling(allDownloadMetrics.Count * _config.Quantile);
+
+                if (completedDownloads.Count >= requiredCompletions && activeDownloads.Count > 0)
```
eric-wang-1990, line 427: **Tracing formula disagrees with actual detection**

The `requiredCompletions` used here for the trace gate:
```csharp
int requiredCompletions = (int)Math.Ceiling(allDownloadMetrics.Count * _config.Quantile);
```
only counts active downloads, but the internal `IdentifyStragglerDownloads` overload that actually runs detection uses:
```csharp
int totalDownloads = allDownloadMetrics.Count + _completedDownloadMetrics.Count;
int requiredCompletions = (int)Math.Ceiling(totalDownloads * _config.Quantile);
```

Once downloads start completing and moving to `_completedDownloadMetrics`, the tracing threshold here will be lower than the real one. The `straggler_detection_check` event can fire (or not fire) in conditions that do not match what the detector actually did, making it misleading to debug against.

Fix: use `(_activeDownloadMetrics.Count + _completedDownloadMetrics.Count)` here too, or pass `requiredCompletions` back from the internal method so the trace always reflects the real threshold.

eric-wang-1990 (approved):

csharp/src/Reader/CloudFetch/StragglerDownload/StragglerDownloadDetector.cs
msrathore-db: Fixed in 88a8ba9 — the tracing block now mirrors the deduped formula from the internal detection overload (`HashSet` of active offsets, `completedFromHistory` filtered to entries not also in active, `totalDownloads = active + completedFromHistory.Count`). The trace gate now matches actual detection's gate, so the `straggler_detection_check` event fires under the same condition.

csharp/src/Reader/CloudFetch/CloudFetchDownloader.cs
msrathore-db: Reverting this change in 9352eee — restored the original `ConcurrentDictionary<Task, IDownloadResult>` and `TryRemove(t, ...)`. On reflection, the original bookkeeping bug (TryRemove using the antecedent task while the dict is keyed by the continuation) is silent and harmless — the dict is local to one query, entries are GC'd at method exit, and `Task.WhenAll` over already-completed tasks is essentially free. It's not straggler-related and bundling it here was scope creep.

Will file a separate cleanup PR off main for the bookkeeping fix so it can be reviewed on its own merits.

PR #218, opened 2026-02-05
ci: enable Rust workflow

lidavidm:


lidavidm:
I went ahead and made tests pass.

PR #220, opened 2026-02-05
feat(csharp): override DriverName to return \"ADBC Databricks Driver\"

eric-wang-1990:
## What's Changed

Overrides the virtual `DriverName` property from `SparkConnection` to properly identify the driver as "ADBC Databricks Driver" instead of "ADBC Spark Driver".

This is a cleaner implementation of PR #125, made possible by PR #179's submodule update which included the virtual DriverName support from apache/arrow-adbc PR #3893.

### Changes
- Override `DriverName` property in `DatabricksConnection` (single line!)
- Add unit test verifying `DriverName` returns "ADBC Databricks Driver" and `DriverArrowVersion` returns "1.0.0" via GetInfo API

### Why this is cleaner than the original PR #125
- **Original PR #125**: Required `private new const string DriverName` + `protected override string InfoDriverName` 
- **This PR**: Just `protected override string DriverName` (virtual property pattern)
- **Reason**: The base class now supports virtual `DriverName` property thanks to apache/arrow-adbc PR #3893

### Test Plan
- ✅ Unit test passes: `GetInfo_DriverInformation_ReturnsCorrectValues`
- ✅ Verifies GetInfo API returns correct driver name
- ✅ All existing tests should continue to pass

### DIFF
--- a/csharp/src/DatabricksConnection.cs
+++ b/csharp/src/DatabricksConnection.cs
@@ -101,8 +101,6 @@ namespace AdbcDrivers.Databricks
         // Default namespace
         private TNamespace? _defaultNamespace;
 
-        private HttpClient? _authHttpClient;
-
         /// <summary>
         /// RecyclableMemoryStreamManager for LZ4 decompression.
         /// If provided by Database, this is shared across all connections for optimal pooling.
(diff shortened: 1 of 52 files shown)


csharp/src/DatabricksConnection.cs
```diff
@@ -527,6 +529,9 @@ internal override IArrowArrayStream NewReader<T>(T statement, Schema schema, IRe
             }
 
             HttpClient httpClient = new HttpClient(HiveServer2TlsImpl.NewHttpClientHandler(TlsOptions, _proxyConfigurator));
+            // Set user-agent for reader HTTP client (used for CloudFetch and result fetching)
```
eric-wang-1990: I do not think this is needed?

eric-wang-1990: ✅ Done - Removed the redundant comment. The code is self-explanatory.

eric-wang-1990: (Updated) Removed the entire user-agent setting code block - both the comment and the two lines that set the user-agent. The reader HTTP client will now use the default .NET user-agent.

PR #241, opened 2026-02-20
refactor(csharp): replace manual telemetry models with proto classes

jadewang-db:
- Add sql_driver_telemetry.proto from universe with C# namespace
- Configure protobuf code generation in csproj using Grpc.Tools
- Add Google.Protobuf package dependency
- Replace manual model classes with proto-generated types:
  - TelemetryEvent → OssSqlDriverTelemetryLog
  - DriverSystemConfiguration, SqlExecutionEvent, DriverErrorInfo
- Add ProtoJsonConverter for System.Text.Json integration
- Use JsonFormatter.Default for proto JSON serialization
- Update FrontendLogEntry to reference proto type
- Update tests to use proto types
- Add TelemetryProtoAlignmentTests for proto schema validation

## What's Changed

Closes #NNN.

### DIFF
--- a/.github/workflows/rust_coverage_report.yaml
+++ b/.github/workflows/rust_coverage_report.yaml
@@ -1,0 +1,65 @@
+# Copyright (c) 2025-2026 ADBC Drivers Contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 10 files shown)


samikshya-db (commented):
maybe I am missing something but - I do not understand the need for this PR. There is no `alignment` as such as there is no sync. Can we not work with the existed models that we have ? Also, there are a couple of fields added in this PR but not populated yet

PR #296, opened 2026-03-04
feat(csharp): implement telemetry client with batching and periodic flush

jadewang-db:
## Summary
- Define `ITelemetryClient` interface for batched telemetry export
- Implement `TelemetryClient` with batching queue and periodic flush
- Implement `TelemetryClientManager` singleton with per-host client sharing and reference counting

**Stack**: PR 2/5 — depends on #295


jadewang-db:
Closing as stale — superseded by newer telemetry work.

PR #308, opened 2026-03-05
feat(csharp): wire telemetry into connection/statement lifecycle with E2E tests

jadewang-db:
## 🥞 Stacked PR
Use this [link](https://github.com/adbc-drivers/databricks/pull/308/files) to review incremental changes.
- [**stack/itr5-pr-e2e-tests**](https://github.com/adbc-drivers/databricks/pull/308) [[Files changed](https://github.com/adbc-drivers/databricks/pull/308/files)]
  - stack/fix-telemetry-gaps-design

---------
## Summary

- **Connection lifecycle**: Wire telemetry initialization into `DatabricksConnection` (feature flag check, per-host client management, session context creation) and graceful disposal with reference counting
- **Statement instrumentation**: Override `ExecuteQuery`, `ExecuteQueryAsync`, `ExecuteUpdate`, and `ExecuteUpdateAsync` in `DatabricksStatement` to capture telemetry context (statement type, result format, compression, statement ID, errors) and emit one telemetry event per statement execution
- **Activity tags**: Add telemetry-specific tags across connection, polling, composite reader, and CloudFetch components (poll count, result format, chunk count, bytes downloaded, compression)
- **HttpClientFactory**: Add `CreateTelemetryHttpClient` using full auth/proxy handler chain for telemetry export
- **CapturingTelemetryExporter**: Test double that captures exported telemetry logs for assertion without hitting real endpoints
- **E2E tests**: Comprehensive tests validating all telemetry fields match JDBC driver output, including snake_case proto serialization, statement ID capture, error handling, and performance overhead

## Test Plan

- [x] Unit tests for Activity tag definitions
- [x] E2E tests validating all telemetry fields per design doc
- [x] E2E tests for connection lifecycle (init, dispose, multi-connection sharing)
- [x] E2E tests for statement telemetry (query, update, error paths)
- [x] snake_case field name assertions matching proto wire format
- [x] Performance overhead test
- [x] Build passes with 0 warnings, 0 errors (net8.0)

### DIFF
--- a/csharp/src/DatabricksConnection.cs
+++ b/csharp/src/DatabricksConnection.cs
@@ -105,6 +105,9 @@ namespace AdbcDrivers.Databricks
         // Default namespace
         private TNamespace? _defaultNamespace;
 
+        // Shared OAuth token provider for connection-wide token caching
+        private OAuthClientCredentialsProvider? _oauthTokenProvider;
+
         // Telemetry fields
         private ITelemetryClient? _telemetryClient;
         private string? _host;
@@ -411,8 +414,9 @@ namespace AdbcDrivers.Databricks
                 AddThriftErrorHandler = true
             };
 
-            var result = HttpHandlerFactory.CreateHandlers(config);
-            return result;
+            var result = HttpHandlerFactory.CreateHandlersWithTokenProvider(config);
+            _oauthTokenProvider = result.TokenProvider;
+            return result.Handler;
         }
 
         protected override bool GetObjectsPatternsRequireLowerCase => true;
@@ -630,8 +634,8 @@ namespace AdbcDrivers.Databricks
                     return;
                 }
 
-                // Create HTTP client for telemetry export
-                HttpClient telemetryHttpClient = HttpClientFactory.CreateTelemetryHttpClient(Properties, _host, s_assemblyVersion);
+                // Create HTTP client for telemetry export, reusing the connection's OAuth token provider
+                HttpClient telemetryHttpClient = HttpClientFactory.CreateTelemetryHttpClient(Properties, _host, s_assemblyVersion, _oauthTokenProvider);
 
                 // Get or create telemetry client from manager (per-host singleton)
                 _telemetryClient = TelemetryClientManager.GetInstance().GetOrCreateClient(
--- a/csharp/src/Http/HttpClientFactory.cs
+++ b/csharp/src/Http/HttpClientFactory.cs
@@ -121,16 +121,19 @@ namespace AdbcDrivers.Databricks.Http
         /// <param name="properties">Connection properties.</param>
         /// <param name="host">The Databricks host (without protocol).</param>
         /// <param name="assemblyVersion">The driver version for the User-Agent.</param>
+        /// <param name="existingTokenProvider">Optional existing OAuthClientCredentialsProvider to reuse for token caching.</param>
         /// <returns>Configured HttpClient for telemetry.</returns>
         public static HttpClient CreateTelemetryHttpClient(
             IReadOnlyDictionary<string, string> properties,
             string host,
-            string assemblyVersion)
+            string assemblyVersion,
+            Auth.OAuthClientCredentialsProvider? existingTokenProvider = null)
         {
             const int DefaultTelemetryTimeoutSeconds = 10;
 
             // Use the same auth handler chain as feature flags (PAT, OAuth, WIF, proxy)
-            var handler = HttpHandlerFactory.CreateFeatureFlagHandler(properties, host, DefaultTelemetryTimeoutSeconds);
+            // Reuse existing token provider to avoid duplicate OAuth token fetches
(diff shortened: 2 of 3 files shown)


jadewang-db:
**[High] Static mutable `ExporterOverride` on singleton is a test-safety concern**

`csharp/src/Telemetry/TelemetryClientManager.cs:55`

`ExporterOverride` is a `static` mutable field on the singleton `TelemetryClientManager`. This means:
1. If xUnit runs test classes in parallel, tests will interfere with each other (one test's `SetupCapturingExporter()` will affect another's)
2. Production code now carries test-only infrastructure

**Suggestion:** Consider one of:
- Use a non-static instance-level property and inject a test-specific `TelemetryClientManager` instance (you already have the `internal TelemetryClientManager(bool forTesting)` constructor)
- At minimum, make the E2E test collection serialized with `[Collection("Telemetry")]` to prevent parallel execution conflicts
- Document thread-safety expectations in the property doc comment

---

jadewang-db:
**[Medium] `DatabricksConnectionTelemetryTests` are smoke tests, not telemetry tests**

`csharp/test/E2E/DatabricksConnectionTelemetryTests.cs`

Most tests in this file (e.g., `OpenAsync_InitializesTelemetry`, `Dispose_ReleasesTelemetryClient`, `MultipleConnections_SameHost_ShareTelemetryClient`) only verify that the connection opens and queries execute successfully. They don't assert any telemetry behavior — they just confirm "telemetry didn't break the connection."

Consider either:
- Renaming them to reflect their actual purpose (smoke/integration tests)
- Adding a `CapturingTelemetryExporter` to actually assert telemetry was produced/flushed (like the tests in `ClientTelemetryE2ETests.cs` do)

---

jadewang-db:
**[Low] Potential null reference in `BuildDriverConnectionParams`**

`csharp/src/DatabricksConnection.cs:754`

```csharp
HostUrl = $"https://{_host}:443",
```

If `_host` is null (e.g., `GetHost()` returned null), this produces `"https://:443"`. While `_host` is set earlier in `InitializeTelemetryAsync`, a defensive null check would be safer since `BuildDriverConnectionParams` is called from the same method and `_host` could theoretically be null if `GetHost()` fails silently.

---

csharp/src/Reader/DatabricksOperationStatusPoller.cs
```diff
@@ -77,15 +83,32 @@ public void Start(CancellationToken externalToken = default)
...
                 var request = new TGetOperationStatusReq(operationHandle);
                 var response = await _statement.Client.GetOperationStatus(request, GetOperationStatusTimeoutToken);
+                long pollEndMs = stopwatch.ElapsedMilliseconds;
+
+                // Track telemetry
+                _pollCount++;
+                long pollLatency = pollEndMs - pollStartMs;
+                _totalLatencyMs += pollLatency;
+
+                // Track time to first byte (first successful poll)
```
eric-wang-1990: This records the time for the first poll operation status finish, why do we wanna track this? I do not see how it is related to when first byte arrives

csharp/src/DatabricksConnection.cs
```diff
@@ -426,6 +434,12 @@ internal override IArrowArrayStream NewReader<T>(T statement, Schema schema, IRe
                 isLz4Compressed = metadataResp.Lz4Compressed;
             }
 
+            // Capture statement ID from server response for telemetry
+            if (response.OperationHandle?.OperationId?.Guid != null)
+            {
+                databricksStatement.LastStatementId = new Guid(response.OperationHandle.OperationId.Guid).ToString();
```
eric-wang-1990: Why is this called LastStatementId? Is there a CurrentStatementId?

csharp/src/DatabricksConnection.cs
```diff
@@ -564,10 +587,180 @@ protected override async Task HandleOpenSessionResponse(TOpenSessionResp? sessio
...
+                    isAuthenticated,
+                    telemetryConfig);
+
+                // Create session-level telemetry context for V3 direct-object pipeline
+                TelemetrySession = new TelemetrySessionContext
+                {
+                    SessionId = SessionHandle?.SessionId?.Guid != null
+                        ? new Guid(SessionHandle.SessionId.Guid).ToString()
+                        : null,
+                    AuthType = isAuthenticated ? "token" : "none",
```
eric-wang-1990: Add TODO needs to update AuthType

csharp/src/Reader/DatabricksOperationStatusPoller.cs
jadewang-db: ✅ Done - You're right, tracking "first poll completed" time is not the same as "time to first byte" (which is when the first result data arrives at the client). Removed the misleading `firstPollCompleted` tracking and the per-poll latency accumulation. Kept only the `_pollCount` metric which is straightforward. Proper time-to-first-byte tracking should be done at the reader level where data actually arrives.

---

PR #315, opened 2026-03-07
ci(ci): trigger REST integration tests when SEA path is touched [PECO-2945]

eric-wang-1990:
Detect changes under csharp/src/StatementExecution/ and dispatch a second adbc-csharp-pr-test event with databricks_protocol=rest. Applies to both PR-label and merge queue. Thrift tests unaffected. Closes PECO-2945


msrathore-db:
The cloudFetch tests should be triggered for SEA and thrift since they share the same path.

PR #317, opened 2026-03-07
fix(csharp): use consistent user agent for SEA and Thrift modes

eric-wang-1990:
## Summary

- SEA mode was using a hardcoded `DatabricksJDBCDriverOSS/{version} (ADBC)` user agent
- Now uses `UserAgentHelper.GetUserAgent` — the same as Thrift mode — producing `AdbcDatabricksDriver/{version}`
- Also respects the `adbc.spark.user_agent_entry` property for custom suffixes, consistent with Thrift behavior


eric-wang-1990:
Superseded by #503. The substantive goals here — drop the `DatabricksJDBCDriverOSS` name, honor `adbc.spark.user_agent_entry`, and use a consistent prefix — have all landed on main, and #503 finishes the job by unifying every builder on `UserAgentHelper` with a single capital-`ADBC` prefix sourced from `DatabricksDriverName`. Closing this in favor of #503.

PR #282, opened 2026-03-02
feat(csharp): add connection parameter support — EnablePKFK, EnableMultipleCatalogSupport

msrathore-db:
## 🥞 Stacked PR
Use this [link](https://github.com/adbc-drivers/databricks/pull/282/files/6c4819e..680f8e3) to review incremental changes only.

- [feat/sea-metadata-part1](https://github.com/adbc-drivers/databricks/pull/257) [[Files changed](https://github.com/adbc-drivers/databricks/pull/257/files)]
  - [feat/sea-metadata-part2](https://github.com/adbc-drivers/databricks/pull/258) [[Files changed](https://github.com/adbc-drivers/databricks/pull/258/files/3c9e2bf..4b3f22b)]
    - [feat/sea-metadata-part3](https://github.com/adbc-drivers/databricks/pull/259) [[Files changed](https://github.com/adbc-drivers/databricks/pull/259/files/4b3f22b..3e07fa0)]
      - [feat/sea-metadata-part4](https://github.com/adbc-drivers/databricks/pull/260) [[Files changed](https://github.com/adbc-drivers/databricks/pull/260/files/3e07fa0..cd97cce)]
        - [feat/sea-metadata-part5](https://github.com/adbc-drivers/databricks/pull/261) [[Files changed](https://github.com/adbc-drivers/databricks/pull/261/files/cd97cce..6c4819e)]
          - [feat/sea-metadata-part6](https://github.com/adbc-drivers/databricks/pull/282) [[Files changed](https://github.com/adbc-drivers/databricks/pull/282/files/6c4819e..680f8e3)]

### EnablePKFK
- Read `DatabricksParameters.EnablePKFK` from connection properties (default: true)
- Gate GetPrimaryKeys/GetCrossReference with `MetadataUtilities.ShouldReturnEmptyPKFKResult`
- Remove bare catch blocks — errors propagate to caller (matches Thrift)

### EnableMultipleCatalogSupport
- Read `DatabricksParameters.EnableMultipleCatalogSupport` (default: true)
- **Enabled**: resolve catalog via NormalizeSparkCatalog → connection default
- **Disabled**: GetCatalogs returns ["SPARK"]; GetSchemas/Tables/Columns return empty for explicit non-SPARK catalog; null/SPARK resolves via `SELECT CURRENT_CATALOG()`
- `ResolveEffectiveCatalog` centralizes catalog resolution
- `EffectiveCatalog` computed once per method to avoid repeated SQL calls

### GetSchemas column parsing fix
[...]

PR #325, opened 2026-03-09
fix(csharp): preserve schema for empty result sets in SEA mode (PECO-2949)

eric-wang-1990:
## Problem

When executing a query against an empty table via the Statement Execution API (SEA / REST mode), the `ResultManifest` always contains column schema even when `totalRowCount` is 0. However, `CreateReader` fell through to `EmptyArrowArrayStream` which always returned `new Schema.Builder().Build()` (zero fields), silently discarding the manifest schema.

This means callers of `IArrowArrayStream.Schema` received a schema with no columns for any empty-table query in SEA mode, making it impossible to inspect column metadata without actually having data rows.

## Root Cause

`EmptyArrowArrayStream` was schema-unaware — it hardcoded an empty schema regardless of context. The `CreateReader` branching logic only extracted schema from the manifest in the CloudFetch and inline-data paths, not in the empty-result fallback path.

## Fix

Follow the JDBC driver pattern where `DatabricksResultSetMetaData` is always constructed from `ResultManifest` independently of data presence:

- Refactor `GetSchemaFromManifest` into a null-returning `TryGetSchemaFromManifest` helper
- In the no-data branch of `CreateReader`, extract schema from manifest and pass it to `EmptyArrowArrayStream`
- Make `EmptyArrowArrayStream` accept an optional `Schema` parameter (defaults to empty schema for the null-manifest / DDL case)

## Tests

Added `StatementExecutionEmptyResultSchemaTests` with 3 unit tests:
- `ExecuteQuery_EmptyTable_SchemaContainsCorrectColumns` — column names preserved for empty table
- `ExecuteQuery_EmptyTable_ArrowTypesAreMappedCorrectly` — INT/BIGINT/STRING/BOOLEAN/DOUBLE/DATE/TIMESTAMP all map correctly
- `ExecuteQuery_NullManifest_ReturnsEmptySchema` — null manifest (DDL) still returns empty schema without error

## Known CI Failure (Pre-existing)

`TelemetryTests.CanEnableFileTracingExporterViaEnvVariable(exporterName: "adbcfile")` fails with the `c078a8ec` version of the hiveserver2 submodule (current `main` baseline). This failure is **unrelated to this PR** — the same test passes in PR #282 which updates the submodule to `e42efb47`. Once PR #282 merges, this failure will be resolved.

The failure is caused by a bug in `TelemetryTests.cs` line 69: `Assert.True(string.IsNullOrEmpty(tc.ActivitySourceName))` should be `Assert.False`. The assertion exception is swallowed by the catch block, leaving `activitySourceName` as `""` and the trace file unclosed/unflushed before the length check.

Closes PECO-2949

### DIFF
--- a/csharp/src/StatementExecution/StatementExecutionStatement.cs
+++ b/csharp/src/StatementExecution/StatementExecutionStatement.cs
@@ -405,8 +405,13 @@ namespace AdbcDrivers.Databricks.StatementExecution
             }
             else
             {
-                // No inline data - return empty reader
-                return new EmptyArrowArrayStream();
+                // No data rows, but the manifest contains schema information.
+                // Preserve the schema so callers get correct column metadata even
+                // when the queried table is empty — following the same pattern as
+                // the JDBC driver where ResultManifest schema is always extracted
+                // independently of data presence.
+                Schema schema = TryGetSchemaFromManifest(response.Manifest) ?? new Schema.Builder().Build();
+                return new EmptyArrowArrayStream(schema);
             }
         }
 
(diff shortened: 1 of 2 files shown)


eric-wang-1990:
## CI Failure Analysis

The failing test `TelemetryTests.CanEnableFileTracingExporterViaEnvVariable(exporterName: "adbcfile")` is **unrelated to this PR's changes**. My changes only touch `StatementExecutionStatement.cs` (schema extraction logic) and add `StatementExecutionEmptyResultSchemaTests.cs`.

### Root causes in the test (both pre-existing)

**1. Bug in `TelemetryTests.cs` line 69** (`csharp/hiveserver2/csharp/test/Common/TelemetryTests.cs`):
```csharp
// Current (wrong): asserts ActivitySourceName IS empty, but message says non-empty
Assert.True(string.IsNullOrEmpty(tc.ActivitySourceName), "expecting non-empty ActivitySourceName");
activitySourceName = tc.ActivitySourceName;  // Never reached — exception is caught
```
Should be `Assert.False`. Because the assertion throws, `activitySourceName` stays `""` and line 86 (`Assert.StartsWith(activitySourceName, files.First().Name)`) trivially passes for any filename.

**2. Bug in `TracingFile.cs` `WriteSingleLineAsync`** (arrow-adbc submodule):
```csharp
await stream.CopyToAsync(_currentFileStream);
await stream.FlushAsync();  // Flushes the source MemoryStream, not _currentFileStream!
```
The file stream is never explicitly flushed after writes — only on `Dispose()`. In CI this may produce a 0-byte file if dispose/flush ordering is affected by the test assertion bug above.

Neither issue is introduced by this PR. The `TelemetryTests.cs` in the hiveserver2 submodule was last modified in `7c90d2d` ("move source and test up one folder") and has no prior "Tests Workflow" runs in this repo to compare against.

---

PR #332, opened 2026-03-11
fix(csharp): ensure telemetry flush waits for in-flight operations on dispose

jadewang-db:
## Summary

Fix telemetry E2E test failures caused by three issues:

- **OAuth auth HTTP client**: `CreateFeatureFlagHandler` skipped creating the auth HTTP client when an existing token provider was passed, causing `InitializeTelemetry` to throw silently
- **Singleton concurrency**: `ExporterOverride` was a plain static, leaking across concurrent test classes. Replaced with `AsyncLocal<T>` for per-async-context isolation
- **Cache bypass**: When `ExporterOverride` is set, bypass the per-host TelemetryClient singleton cache so each test connection gets its own client with the capturing exporter

## Test plan

- [x] Telemetry_QueryExecution_CapturesAllFields passes
- [x] Telemetry_ErrorQuery_CapturesErrorInfo passes
- [x] Telemetry_FeatureFlagDisabled_NoExport passes (no cross-test pollution)
- [x] All other telemetry E2E tests unaffected

### DIFF
--- a/csharp/src/DatabricksConnection.cs
+++ b/csharp/src/DatabricksConnection.cs
@@ -112,7 +112,6 @@ namespace AdbcDrivers.Databricks
         private ITelemetryClient? _telemetryClient;
         private string? _host;
         internal TelemetrySessionContext? TelemetrySession { get; private set; }
-        internal string? TelemetryInitError { get; private set; }
 
         /// <summary>
         /// RecyclableMemoryStreamManager for LZ4 decompression.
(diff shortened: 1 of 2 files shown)


csharp/src/Telemetry/TelemetryClient.cs
```diff
@@ -209,7 +214,7 @@ public void Enqueue(TelemetryFrontendLog log)
         /// driver functionality.
         /// </para>
         /// </remarks>
-        public async Task FlushAsync(CancellationToken ct = default)
+        public async Task FlushAsync(CancellationToken ct = default, bool waitForSemaphore = false)
```
jadewang-db: let's revert change on this, this is not the real cause of the test failures

jadewang-db: ✅ Done — reverted the `waitForSemaphore` parameter from `FlushAsync`. Restored the original non-blocking semaphore behavior across `TelemetryClient`, `ITelemetryClient` interface, `DatabricksConnection`, and the mock in unit tests.

---

PR #336, opened 2026-03-11
fix(csharp): use TypeText for Spark:DataType:SqlName in SEA schema metadata

eric-wang-1990:
## Summary
- `TypeName` is an enum of base types only (`ARRAY`, `DECIMAL`, `STRUCT`) and loses detail for complex and parameterized types
- `TypeText` contains the full SQL type specification (`ARRAY<INT>`, `DECIMAL(10,2)`, `STRUCT<id:INT,name:STRING>`) — matching what the Thrift server embeds in `Spark:DataType:SqlName`
- Falls back to `TypeName` if `TypeText` is absent

## Test plan
- [ ] Verify `Spark:DataType:SqlName` for a `DECIMAL(10,2)` column returns `DECIMAL(10,2)` not `DECIMAL`
- [ ] Verify `Spark:DataType:SqlName` for an `ARRAY<INT>` column returns `ARRAY<INT>` not `ARRAY`
- [ ] Verify `Spark:DataType:SqlName` for a `STRUCT<id:INT,name:STRING>` column returns the full struct spec

### DIFF
--- a/csharp/src/StatementExecution/StatementExecutionStatement.cs
+++ b/csharp/src/StatementExecution/StatementExecutionStatement.cs
@@ -482,7 +482,7 @@ namespace AdbcDrivers.Databricks.StatementExecution
                 // it here for now. Add it if a consumer requires it (PECO-2950).
                 var metadata = new Dictionary<string, string>
                 {
-                    ["Spark:DataType:SqlName"] = ColumnMetadataHelper.GetSparkSqlName(column.TypeName ?? string.Empty)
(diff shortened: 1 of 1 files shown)


eric-wang-1990:
Reverting — GetSparkSqlName calls GetBaseTypeName which strips type parameters regardless of input, so passing TypeText instead of TypeName doesn't actually preserve detail for complex/parameterized types. A proper fix needs to update GetSparkSqlName to return TypeText directly rather than routing through the base-type parser.

PR #337, opened 2026-03-12
fix(csharp): normalize TypeName once in TryGetSchemaFromManifest

eric-wang-1990:
PR #330 follow-up: the two uses of column.TypeName on lines 485 and 497 were inconsistent — the Arrow type mapping had no null guard while the metadata assignment used `?? string.Empty`. Extract a single `typeName` local to make null handling uniform and remove the silent divergence.

Add two unit tests:
- null TypeName falls back to StringType without throwing
- DECIMAL(10,2), ARRAY, MAP, STRUCT are all mapped without exception

## What's Changed

Closes #NNN.

### DIFF
--- a/csharp/test/Unit/StatementExecution/StatementExecutionEmptyResultSchemaTests.cs
+++ b/csharp/test/Unit/StatementExecution/StatementExecutionEmptyResultSchemaTests.cs
@@ -250,49 +250,6 @@ namespace AdbcDrivers.Databricks.Tests.Unit.StatementExecution
         }
 
         [Fact]
-        public async Task ExecuteQuery_NullTypeName_TreatedAsUnknownType()
-        {
-            // Arrange: a column with a null TypeName must not throw — it falls back to StringType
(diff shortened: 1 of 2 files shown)


csharp/test/Unit/StatementExecution/StatementExecutionEmptyResultSchemaTests.cs
```diff
@@ -249,6 +249,87 @@ public async Task ExecuteQuery_EmptyTable_SqlNameAliasesNormalized()
             Assert.Equal("SMALLINT", fields[2].Metadata["Spark:DataType:SqlName"]);
         }
 
+        [Fact]
+        public async Task ExecuteQuery_EmptyTable_ComplexAndDecimalTypesAreMappedCorrectly()
```
eric-wang-1990: Why need add these, we already have csharp/test/ColumnMetadataHelperTests.cs which test all type mapping?

msrathore-db (approved):
Thanks

PR #347, opened 2026-03-13
fix(csharp): prevent infinite polling in SEA metadata queries

msrathore-db:
## Summary
- **Bug**: `ExecuteMetadataSqlAsync` creates internal statements with `_queryTimeoutSeconds=0` (no timeout). When called via `ExecuteQuery()` on a metadata statement (e.g. `GetTables`, `GetColumns`, `GetCatalogs`), `CancellationToken.None` is passed through, causing `PollUntilCompleteAsync`'s `while(true)` loop to run indefinitely if the server query doesn't complete within the initial 10s REST API wait.
- **Impact**: PowerBI users see frozen UI / infinite spinners when browsing tables or columns via the SEA (REST) protocol, with no way to cancel. The only escape is killing the process.
- **Fix**: Wrap all metadata SQL execution with `CreateMetadataTimeoutCts()` (using `_waitTimeoutSeconds`, default 10s), matching the pattern already used by `GetObjects()`. Links caller-provided tokens so either the caller or the metadata timeout can terminate the operation.
- **Why Thrift is unaffected**: Thrift uses native `TGetTablesReq` RPC — a single request-response call with no SQL or polling.

### Root cause details

The code path: `ExecuteQuery()` → `ExecuteQueryAsync(CancellationToken.None)` → `ExecuteMetadataCommandAsync` → `GetTablesAsync(CancellationToken.None)` → `ExecuteMetadataSqlAsync(sql, CancellationToken.None)` → creates new `StatementExecutionStatement` with `_queryTimeoutSeconds=0` → `PollWithTimeoutAsync` sees `<= 0` → `PollUntilCompleteAsync(statementId, CancellationToken.None)` → infinite `while(true)`.

All SEA metadata commands are affected: `GetCatalogs`, `GetSchemas`, `GetTables`, `GetColumns`, `GetColumnsExtended`, `GetPrimaryKeys`, `GetCrossReference`, and `GetCurrentCatalog()` (`SELECT CURRENT_CATALOG()`).

## Test plan
- [x] All 631 unit tests pass
- [x] All 18 SEA metadata E2E tests pass (`SeaMetadataE2ETests`)
- [ ] Manual PowerBI validation with SEA protocol against slow/cold warehouse

This pull request was AI-assisted by Isaac.


csharp/src/StatementExecution/StatementExecutionConnection.cs
```diff
@@ -734,16 +734,28 @@ async Task IGetObjectsDataProvider.PopulateColumnInfoAsync(string? catalogPatter
 
         internal async Task<List<RecordBatch>> ExecuteMetadataSqlAsync(string sql, CancellationToken cancellationToken = default)
         {
+            // Ensure metadata SQL always has a timeout to prevent infinite polling.
+            // The caller (e.g. GetObjects) may already supply a timeout token, but if
+            // the token is CancellationToken.None (e.g. when called from
+            // ExecuteMetadataCommandAsync → GetTablesAsync via ExecuteQuery()), the
+            // inner statement would have _queryTimeoutSeconds=0 and poll forever.
+            // Link the caller's token with a metadata timeout as a safety net.
+            using var metadataTimeoutCts = CreateMetadataTimeoutCts();
```
eric-wang-1990: It seems we have same bug for Thrift, can we fix that too?

PR #378, opened 2026-03-31
feat(rust): add prepared statement and parameter binding support

vikrantpuppala:
## Summary
- Add prepared statement and parameter binding support to the Rust ADBC driver via the SEA API's positional parameters
- Implement `prepare()`, `bind()`, `bind_stream()`, `get_parameter_schema()`, and fix `execute_schema()` to use `DESCRIBE QUERY` instead of full execution
- Add SQL `?` marker parsing (ignoring string literals/comments), Arrow-to-Databricks type conversion, and value serialization
- `bind_stream()` executes one SEA API call per row (no batch INSERT optimization)
- Fix missing `databricks.auth.type` in integration and FFI test setup

### Files changed
| File | Change |
|---|---|
| `src/types/sea.rs` | `StatementParameter` struct, `parameters` field on `ExecuteStatementRequest` and `ExecuteParams` |
| `src/statement/mod.rs` | New fields, `prepare`/`bind`/`bind_stream`/`get_parameter_schema` impl, `execute_schema` fix, updated `execute` |
| `src/statement/params.rs` (new) | SQL parser, Arrow→SEA conversion, DESCRIBE QUERY parsing, type mappings (30 unit tests) |
| `src/client/sea.rs` | Wire `parameters` from `ExecuteParams` to `ExecuteStatementRequest` |
| `src/reader/mod.rs` | `VecBatchReader` + `ResultReaderAdapter::from_batches()` for stream result collection |
| `tests/prepared_statement_e2e.rs` (new) | 10 E2E tests: parameterized SELECT, multiple params, NULL, boolean, bigint, bind_stream, execute_schema, rebind |
| `tests/integration.rs`, `tests/ffi_round_trip.rs` | Add missing `databricks.auth.type` option |

## Test plan
- [x] 30 unit tests for SQL parsing, Arrow→SEA conversion, type mapping, DESCRIBE QUERY parsing
- [x] 10 E2E tests passing against dogfood (parameterized queries, bind_stream, execute_schema via DESCRIBE QUERY, rebind/re-execute)
- [x] All 286 existing unit tests pass
- [x] `cargo fmt`, `cargo clippy` clean

This pull request was AI-assisted by Isaac.

PR #380, opened 2026-04-03
fix(rust): implement StatementPrepare as no-op for ADBC consumer compat

jatorre:
## Summary

- DuckDB's `adbc_scanner` (and likely other ADBC consumers) calls `StatementPrepare` before `StatementExecuteQuery`. The current stub returns `NOT_IMPLEMENTED`, which breaks `adbc_scan`.
- Databricks doesn't have server-side prepared statements, so implement `prepare` as a no-op that validates a query has been set via `SetSqlQuery`. This matches the Go driver's behavior.

## Test plan

- [ ] Existing unit/integration tests pass (`cargo test`)
- [ ] Manual test: load driver through DuckDB's `adbc_scanner` and run `adbc_scan(conn, 'SELECT 1')`

### DIFF
--- a/rust/src/statement.rs
+++ b/rust/src/statement.rs
@@ -105,9 +105,15 @@ impl adbc_core::Statement for Statement {
     }
 
     fn prepare(&mut self) -> Result<()> {
-        Err(DatabricksErrorHelper::not_implemented()
-            .message("prepare")
-            .to_adbc())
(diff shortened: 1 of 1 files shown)


jatorre:
Closing — upstream #378 (merged) implemented `prepare()` with full prepared statement support, which supersedes this no-op fix. Thanks!

PR #383, opened 2026-04-12
feat(go): enable CloudFetch for parallel data downloads

jatorre:
## Summary

- Enable CloudFetch by passing `WithCloudFetch(true)` to databricks-sql-go
- The driver already passes `WithMaxDownloadThreads()` but never enables CloudFetch itself, so all result data flows through inline Thrift IPC (single-threaded)
- This is a one-line fix that matches the Rust and C# drivers, which both use parallel CloudFetch by default

## Benchmark

Tested with DuckDB `adbc_scanner` against Slovenia OSM geospatial data on a warm Databricks SQL Warehouse:

| Dataset | Rows | Without CloudFetch | With CloudFetch | Speedup |
|---------|------|--------------------|-----------------|---------|
| Points (POIs) | 51K | 2m 11s | **44s** | **3x** |
| Polygons (Buildings) | 934K | 4m 30s | **2m 16s** | **2x** |

CPU utilization also increased from 25% to 51%, confirming that parallel downloads are keeping the client busy instead of waiting on a single stream.

## Test plan

- [x] `go build ./...` compiles clean
- [x] Verified with real Databricks SQL Warehouse (PAT auth)
- [x] Benchmarked with geometry and non-geometry workloads
- [ ] Existing CI tests should pass (no behavior change for small inline results; CloudFetch only activates for larger result sets where the server provides external links)


jatorre:
Friendly ping @lidavidm @jadewang-db — tiny 3-line change that turns on CloudFetch by default for the Go driver, matching the behavior other ADBC drivers already ship. Rebased onto current main, CI green. Would love to get this in. Thanks!

lidavidm:
@jadewang-db what is the status of the Rust implementation? Should we focus on that instead?

PR #386, opened 2026-04-14
feat(csharp): bump driver version to 1.1.0

eric-wang-1990:
## Summary

- Updates `hiveserver2` submodule to pick up [adbc-drivers/hiveserver2#32](https://github.com/adbc-drivers/hiveserver2/pull/32), which makes `ProductVersionDefault` a virtual property in `SparkConnection`
- Overrides `ProductVersionDefault` in `DatabricksConnection` to report `1.1.0` in User-Agent headers

## Test plan

- [ ] Verify User-Agent header reports `ADBCDatabricksDriver/1.1.0` in query history

### DIFF
--- a/csharp/Directory.Packages.props
+++ b/csharp/Directory.Packages.props
@@ -25,7 +25,7 @@
     <PackageVersion Include="K4os.Compression.LZ4.Streams" Version="1.3.8" />
     <PackageVersion Include="Microsoft.IO.RecyclableMemoryStream" Version="3.0.1" />
     <!-- Testing -->
-    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="18.0.1" />
+    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="18.3.0" />
     <PackageVersion Include="xunit" Version="2.9.3" />
     <PackageVersion Include="xunit.runner.visualstudio" Version="3.1.5" />
     <PackageVersion Include="Xunit.SkippableFact" Version="1.5.61" />
--- a/csharp/src/DatabricksConnection.cs
+++ b/csharp/src/DatabricksConnection.cs
@@ -442,6 +442,7 @@ namespace AdbcDrivers.Databricks
         protected override bool GetObjectsPatternsRequireLowerCase => true;
 
         protected override string DriverName => DatabricksDriverName;
+        protected override string ProductVersionDefault => "1.1.0";
 
         internal override IArrowArrayStream NewReader<T>(T statement, Schema schema, IResponse response, TGetResultSetMetadataResp? metadataResp = null)
         {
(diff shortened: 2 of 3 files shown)


csharp/src/DatabricksConnection.cs
```diff
@@ -442,6 +443,7 @@ protected override HttpMessageHandler CreateHttpHandler()
         protected override bool GetObjectsPatternsRequireLowerCase => true;
 
         protected override string DriverName => DatabricksDriverName;
+        protected override string ProductVersionDefault => DriverVersion;
```
birschick-bq, line 445: Keep in mind that the version that GetInfo returns is the version from the assembly's ProductVersion field, typically...
https://github.com/adbc-drivers/hiveserver2/blob/0c8f1c857509cee9b572b0698c7ba3a423b58936/csharp/src/Hive2/HiveServer2Connection.cs#L855

eric-wang-1990: Yes, but we only care about the userAgent which get surface to Databricks backend and that is using the ProductVersionDefault: https://github.com/adbc-drivers/hiveserver2/blob/0c8f1c857509cee9b572b0698c7ba3a423b58936/csharp/src/Spark/SparkHttpConnection.cs#L290C69-L290C90

jadewang-db: can we align these 2?

birschick-bq: Understood. Thanks for the explanation.

birschick-bq (approved):

csharp/Directory.Packages.props
```diff
@@ -25,7 +25,7 @@
     <PackageVersion Include="K4os.Compression.LZ4.Streams" Version="1.3.8" />
     <PackageVersion Include="Microsoft.IO.RecyclableMemoryStream" Version="3.0.1" />
     <!-- Testing -->
-    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="18.0.1" />
+    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="18.3.0" />
```
msrathore-db, line 28: Why do we need this?

msrathore-db (approved):

csharp/src/DatabricksConnection.cs
eric-wang-1990: @jadewang-db I think so, @birschick-bq do you see an issue to change GetProductVersion not using the Assembly version(Mashup)?
I think it will be used in the future to get the driver version?

csharp/Directory.Packages.props
eric-wang-1990: Required transitively — the hiveserver2 submodule update pulls in Apache.Arrow.Adbc.Testing, which depends on Microsoft.NET.Test.Sdk >= 18.3.0. Without this bump, the build fails due to the version conflict.

csharp/src/DatabricksConnection.cs
birschick-bq: @eric-wang-1990 

Each driver implementation is free to implement [GetProductVersion](https://github.com/adbc-drivers/hiveserver2/blob/0c8f1c857509cee9b572b0698c7ba3a423b58936/csharp/src/Hive2/HiveServer2Connection.cs#L855). My thought is that you'd want to include an update to the third digit (build) to get the specific version with major/minor indicating specific functional stability.

PR #388, opened 2026-04-16
fix(csharp): catch object-not-found exceptions in SEA metadata

msrathore-db:
## Summary
- **Bug**: SEA metadata queries (GetSchemas, GetTables, GetColumns, etc.) throw `DatabricksException` when the server returns `NO_SUCH_CATALOG_EXCEPTION`, `SCHEMA_NOT_FOUND`, or `TABLE_OR_VIEW_NOT_FOUND`. JDBC catches these and returns empty result sets — ADBC was propagating the exception to the caller.
- **Fix**: Added `IsObjectNotFoundException()` to `DatabricksException` and wrapped all SEA metadata methods with `catch (DatabricksException ex) when (ex.IsObjectNotFoundException())` that returns empty results, matching JDBC behavior.
- **Scope**: Covers all SEA metadata code paths:
  - `StatementExecutionStatement`: GetSchemas, GetTables, GetColumns, GetColumnsExtended (both DESC TABLE and three-call fallback), GetPrimaryKeys, FetchCrossReference
  - `StatementExecutionConnection` (`IGetObjectsDataProvider`): GetSchemas, GetTables, PopulateColumnInfo

This should eliminate all 45 `NO_SUCH_CATALOG_EXCEPTION` errors seen in the JDBC/ADBC comparator, converting them to matches (both return empty).

## Test plan
- [x] All 647 unit tests pass
- [ ] Run JDBC/ADBC metadata comparator and verify NO_SUCH_CATALOG_EXCEPTION errors become empty-result matches
- [ ] Verify valid catalog/schema/table queries still return correct results (not affected by catch)

This pull request was AI-assisted by Isaac.


eric-wang-1990 (approved):
Can we add some tests as well

PR #389, opened 2026-04-18
fix(csharp): add scope=sql to RefreshTokenAsync for AAD service principal tokens

eric-wang-1990:
### Bug fix: `scope=sql` missing in `RefreshTokenAsync`

`TokenRefreshDelegatingHandler` calls `RefreshTokenAsync` when the connection token is near expiry. This sends an AAD token to the Databricks OIDC endpoint (`/oidc/v1/token`) using the `urn:ietf:params:oauth:grant-type:jwt-bearer` grant type.

For AAD **service principal** tokens (no `scp` claim), the server reads the OAuth scope from the HTTP request parameter. `RefreshTokenAsync` was not including `scope=sql`, so the server returned `invalid_scope`. The exception was silently swallowed, `_tokenExchangeAttempted` was set to prevent any retry, and the token was never refreshed. Both the original AAD token and the Databricks token expire around T=60min → customer 401s.

The fix is to add `scope=sql` to the `RefreshTokenAsync` form body, matching what `ExchangeTokenAsync` (used by `MandatoryTokenExchangeDelegatingHandler`) already sends.

Note: this only affects AAD **service principal** (M2M) tokens. AAD user tokens carry an `scp` claim that the server reads directly from the token itself, so they are unaffected.

Root cause traced through `JwtBearerGrantHandler.updateAADSPClaimWithScope` in universe: for SP tokens, `OAuthScopeValidator.validate(None)` → `INVALID_SCOPE` when no scope is in the request.

Fixes: PECO-2977

### Improvement: `TokenRefreshDelegatingHandler` — fire-and-forget → blocking async

Previously the token refresh was started in the background; the request proceeded immediately with the old (near-expiry) token. Changed to `await` the refresh before forwarding the request, so the first request near expiry uses a fresh token. Concurrent requests queue through a `SemaphoreSlim(1,1)` — by the time the second thread enters, the first has already refreshed, so it skips and returns the cached token immediately.

### Improvement: `TokenRefreshDelegatingHandler` — multiple refresh cycles

Previously `_tokenExchangeAttempted` permanently prevented a second refresh regardless of whether the refreshed token itself later neared expiry. Replaced with a `Dictionary<string, string>` mapping each near-expiry token to its refreshed replacement. Once a refreshed token nears expiry, it is refreshed again in a new cycle. On failure, the token is mapped to itself to prevent per-request retries for that specific token.

### Bug fix: `MandatoryTokenExchangeDelegatingHandler` — stale token returned on passthrough

When `TokenRefreshDelegatingHandler` upstream provides a fresh Databricks token, `MandatoryTokenExchangeDelegatingHandler` correctly skips exchange. However, the old code returned `_currentToken ?? bearerToken`, which could be the previously-exchanged (stale) token. Fixed by returning `bearerToken` directly when no exchange is needed.

### Refactor: `MandatoryTokenExchangeDelegatingHandler` — simplify caching

Replaced `_lastSeenToken` / `_pendingExchange` / `_tokenLock` with a `SemaphoreSlim(1,1)` and a `Dictionary<string, string>` mapping each original external token to its exchanged Databricks token. Each distinct external token (e.g. a rotated M2M token from `OAuthDelegatingHandler`) is exchanged independently and cached. On failure, the original token is cached to itself to prevent repeated exchange attempts for that token.

## Test plan

- [x] Updated `RefreshTokenAsync_SendsCorrectRequestFormat` to assert `scope=sql` is present in the request body
- [x] Updated `TokenRefreshDelegatingHandler` tests to reflect blocking async behaviour and multiple refresh cycle support
- [x] Updated `MandatoryTokenExchangeDelegatingHandler` tests to cover per-token caching and rotated external token handling
- [x] Added `SendAsync_WhenUpstreamProvidesFreshDatabricksToken_UsesFreshToken` to cover the stale token passthrough fix
- [ ] Run unit tests locally: `dotnet test csharp/test/Unit`
- [ ] E2E validation: confirm AAD SP token queries survive past 60-minute AAD token expiry

Closes #PECO-2977


msrathore-db (approved):
LGTM

csharp/src/Auth/MandatoryTokenExchangeDelegatingHandler.cs
```diff
@@ -119,7 +121,7 @@ private async Task PerformTokenExchangeIfNeeded(string bearerToken, Cancellation
                 // If this token was already processed (by us or another concurrent request)
                 if (_lastSeenToken == bearerToken)
                 {
-                    return;
+                    return _currentToken ?? bearerToken;
```
jadewang-db: logic here is weired.

csharp/src/Auth/MandatoryTokenExchangeDelegatingHandler.cs
```diff
@@ -80,22 +80,24 @@ private bool NeedsTokenExchange(string bearerToken)
         }
 
         /// <summary>
-        /// Performs token exchange if needed.
+        /// Returns the token to use for the request, performing exchange if needed.
+        /// If the token is already Databricks-issued, returns it directly without exchange.
         /// </summary>
         /// <param name="bearerToken">The bearer token to potentially exchange.</param>
         /// <param name="cancellationToken">A cancellation token.</param>
-        private async Task PerformTokenExchangeIfNeeded(string bearerToken, CancellationToken cancellationToken)
```
jadewang-db, line 87: logic of this class can be simplified to just have a member of exchanged token, and if it's available, we can just use it.

if not, go through the exchange process and update the exchanged token?

csharp/src/Auth/MandatoryTokenExchangeDelegatingHandler.cs
jadewang-db, line 127: maybe just return the token as a return value of this call.

csharp/src/Auth/TokenRefreshDelegatingHandler.cs
```diff
@@ -71,81 +71,63 @@ public TokenRefreshDelegatingHandler(
...
+        private async Task EnsureTokenFreshAsync(CancellationToken cancellationToken)
         {
-            if (!NeedsTokenRenewal())
-            {
-                return;
-            }
-
-            bool needsRenewal;
+            Task? taskToAwait;
             lock (_tokenLock)
```
jadewang-db: use async lock await

csharp/src/Auth/TokenRefreshDelegatingHandler.cs
```diff
@@ -71,81 +71,63 @@ public TokenRefreshDelegatingHandler(
...
+                await taskToAwait;
             }
+        }
 
-            // Start token refresh in the background
-            _pendingTokenTask = Task.Run(async () =>
+        /// <summary>
+        /// Performs the token refresh network call.
+        /// </summary>
+        private async Task DoRefreshAsync(CancellationToken cancellationToken)
```
jadewang-db: if the new token as return value, the logic of this method will be simpler

jadewang-db (changes requested):

csharp/src/Auth/MandatoryTokenExchangeDelegatingHandler.cs
eric-wang-1990: Fixed — the handler now uses a single `_currentToken` field protected by a `SemaphoreSlim`. The weird `_lastSeenToken == bearerToken` branch is gone entirely. On failure, `_currentToken` is set to the original bearer token, so subsequent requests return it directly without retrying.

---

csharp/src/Auth/MandatoryTokenExchangeDelegatingHandler.cs
eric-wang-1990: Done — refactored exactly as suggested. The handler now has a single `_currentToken` (protected by `SemaphoreSlim`). Exchange is attempted once per handler instance: on success `_currentToken` = exchanged Databricks token; on failure `_currentToken` = original bearer token (so subsequent requests skip re-exchanging). Removed `_lastSeenToken`, `_pendingExchange`, and `_tokenLock` entirely.

---

csharp/src/Auth/MandatoryTokenExchangeDelegatingHandler.cs
eric-wang-1990: Done — `EnsureTokenFreshAsync` now returns `Task<string>` and `SendAsync` uses it directly: `string tokenToUse = await EnsureTokenFreshAsync(cancellationToken);` with no separate lock read needed.

---

csharp/src/Auth/TokenRefreshDelegatingHandler.cs
eric-wang-1990: Done — replaced `object _tokenLock` / `Task? _refreshTask` with a `SemaphoreSlim _refreshLock`. `EnsureTokenFreshAsync` now uses `await _refreshLock.WaitAsync(cancellationToken)` as the async lock, serializing refresh attempts without blocking a thread.

---

csharp/src/Auth/TokenRefreshDelegatingHandler.cs
eric-wang-1990: Done — `EnsureTokenFreshAsync` now returns the current token as `Task<string>`, so `SendAsync` is simply `string tokenToUse = await EnsureTokenFreshAsync(cancellationToken);` with no extra lock or field read.

---

csharp/src/Auth/BoundedTokenCache.cs
```diff
@@ -0,0 +1,65 @@
...
+    internal class BoundedTokenCache
+    {
+        private readonly int _maxSize;
+        private readonly Dictionary<string, string> _cache;
+        private readonly Queue<string> _insertionOrder;
+
+        public BoundedTokenCache(int maxSize = 10)
+        {
+            _maxSize = maxSize;
+            _cache = new Dictionary<string, string>(maxSize);
```
jadewang-db: concurrentdictionary?

csharp/src/Auth/MandatoryTokenExchangeDelegatingHandler.cs
```diff
@@ -44,7 +45,7 @@ internal class MandatoryTokenExchangeDelegatingHandler : DelegatingHandler
         // On failure, the original token is mapped to itself to prevent repeated exchange attempts.
         // Keyed by the original bearer token so that when the upstream token rotates
         // (e.g. OAuthDelegatingHandler refreshes the M2M token), the new token is exchanged fresh.
-        private readonly BoundedTokenCache _tokenCache = new BoundedTokenCache();
+        private readonly Dictionary<string, string> _tokenCache = new Dictionary<string, string>();
```
jadewang-db: concurrentdictionary

jadewang-db (approved):

csharp/src/Auth/BoundedTokenCache.cs
eric-wang-1990: Removed `BoundedTokenCache` entirely — we now use a plain `Dictionary<string, string>` directly in each handler. Since all dictionary access happens inside the handler's `SemaphoreSlim(1,1)`, thread safety is already guaranteed by the caller. `ConcurrentDictionary` would add unnecessary overhead for a map that only ever has a handful of entries and is already serialized.

---

csharp/src/Auth/MandatoryTokenExchangeDelegatingHandler.cs
eric-wang-1990: Using a plain `Dictionary<string, string>` — all access happens inside the `SemaphoreSlim(1,1)` so only one thread can read or write at a time. `ConcurrentDictionary` would be redundant here.

---

csharp/src/Auth/BoundedTokenCache.cs
eric-wang-1990: Updated to `ConcurrentDictionary<string, string>` (the `BoundedTokenCache` file has been removed entirely).

---

csharp/src/Auth/MandatoryTokenExchangeDelegatingHandler.cs
eric-wang-1990: Updated to `ConcurrentDictionary<string, string>`.

---

csharp/src/Auth/MandatoryTokenExchangeDelegatingHandler.cs
```diff
@@ -52,7 +57,7 @@ internal class MandatoryTokenExchangeDelegatingHandler : DelegatingHandler
         public MandatoryTokenExchangeDelegatingHandler(
             HttpMessageHandler innerHandler,
             ITokenExchangeClient tokenExchangeClient,
-            string? identityFederationClientId = null)
+            string? identityFederationClientId)
```
eric-wang-1990: Why change this?

csharp/src/Auth/MandatoryTokenExchangeDelegatingHandler.cs
```diff
@@ -80,113 +85,91 @@ private bool NeedsTokenExchange(string bearerToken)
...
-                {
-                    _lastSeenToken = bearerToken;
-                }
-                return;
-            }
-
-            // Wait for any pending exchange to complete first (could be for a different token)
-            Task? exchangeToAwait = null;
-            lock (_tokenLock)
+            await _exchangeLock.WaitAsync(cancellationToken).ConfigureAwait(false);
```
eric-wang-1990, line 107: This means even we already have exchanged the token we need to wait for the lock again? But in theory we can return early without await the lock.

csharp/src/Auth/TokenRefreshDelegatingHandler.cs
```diff
@@ -62,96 +67,80 @@ public TokenRefreshDelegatingHandler(
...
+        private async Task<string> EnsureTokenFreshAsync(CancellationToken cancellationToken)
         {
-            if (!NeedsTokenRenewal())
-            {
-                return;
-            }
-
-            bool needsRenewal;
-            lock (_tokenLock)
+            await _refreshLock.WaitAsync(cancellationToken).ConfigureAwait(false);
```
eric-wang-1990, line 94: This means even we already have exchanged the token we need to wait for the lock again? But in theory we can return early without await the lock.

PR #394, opened 2026-04-21
fix(csharp): fetch SEA inline chunks lazily instead of all at once

eric-wang-1990:
## Summary

- `InlineArrowStreamReader` previously fetched all chunks eagerly in the constructor, concatenating them into a single `byte[]` and buffering the entire result in memory
- Replaced with lazy chunk fetching: only the first chunk (already available inline in the response) is loaded upfront
- Subsequent chunks are fetched via `GetResultChunkAsync` on demand as the consumer reads, keeping memory bounded to one chunk at a time
- Removed the now-unused `CancellationToken` constructor parameter (no async work happens in the constructor anymore)

## Test plan

- [ ] Verify single-chunk inline results still work correctly
- [ ] Verify multi-chunk inline results are fetched lazily (each chunk fetched only when previous is exhausted)
- [ ] Verify LZ4-compressed multi-chunk inline results work correctly
- [ ] Verify dispose releases only the current chunk's stream/reader

### DIFF
--- a/csharp/src/StatementExecution/StatementExecutionStatement.cs
+++ b/csharp/src/StatementExecution/StatementExecutionStatement.cs
@@ -432,7 +432,7 @@ namespace AdbcDrivers.Databricks.StatementExecution
                 // Inline results - may be split across multiple chunks
                 int totalChunks = response.Manifest?.Chunks?.Count ?? 1;
                 return new InlineArrowStreamReader(_client, _currentStatementId!, response.Result.Attachment,
-                    isLz4Compressed, totalChunks, _lz4BufferPool, cancellationToken);
+                    isLz4Compressed, totalChunks, _lz4BufferPool);
             }
             else
             {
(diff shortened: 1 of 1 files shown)


csharp/src/StatementExecution/StatementExecutionStatement.cs
```diff
@@ -700,14 +700,22 @@ public void Dispose()
 
         /// <summary>
         /// Reader for inline results in Arrow IPC stream format.
-        /// Handles both single-chunk and multi-chunk inline results by fetching
-        /// all chunks and concatenating them into a single Arrow stream.
+        /// Fetches chunks lazily one at a time as the consumer reads, keeping memory
+        /// usage bounded to a single chunk regardless of total result size.
         /// Supports LZ4_FRAME compressed data.
         /// </summary>
         private class InlineArrowStreamReader : IArrowArrayStream
```
jadewang-db, line 707: we should have a similar read in thrift, can we use that and remove this one instead?  IIRC, this was introduced for inline where we only have one chunk

eric-wang-1990:
Good question — I looked into reusing `DatabricksReader` but it's tightly coupled to Thrift-specific types (`IHiveServer2Statement`, `TSparkArrowBatch`, `TFetchResultsResp`) and can't be used for SEA directly.

The data granularity also differs fundamentally:
- **Thrift**: `FetchResults` returns a `List<TSparkArrowBatch>` where each batch is a single serialized Arrow record batch
- **SEA inline**: `GetResultChunkAsync` returns a `byte[]` that is a full Arrow IPC stream (potentially containing multiple batches)

The only extractable shared logic would be the LZ4 decompression + `ArrowStreamReader` wrapping (~5 lines in `LoadChunk`), which isn't worth a shared abstraction.

Also confirmed with the author that SEA inline disposition **can** return multiple chunks, so the lazy fetching logic is needed.

---

jadewang-db:
> Good question — I looked into reusing `DatabricksReader` but it's tightly coupled to Thrift-specific types (`IHiveServer2Statement`, `TSparkArrowBatch`, `TFetchResultsResp`) and can't be used for SEA directly.
> 
> The data granularity also differs fundamentally:
> 
> * **Thrift**: `FetchResults` returns a `List<TSparkArrowBatch>` where each batch is a single serialized Arrow record batch
> * **SEA inline**: `GetResultChunkAsync` returns a `byte[]` that is a full Arrow IPC stream (potentially containing multiple batches)
> 
> The only extractable shared logic would be the LZ4 decompression + `ArrowStreamReader` wrapping (~5 lines in `LoadChunk`), which isn't worth a shared abstraction.
> 
> Also confirmed with the author that SEA inline disposition **can** return multiple chunks, so the lazy fetching logic is needed.
> 
> _This comment was generated with [GitHub MCP](http://go/mcps)._

we've done samething for cloudfetch, I think this is the opportunity for us to make this code path protocol indepent

PR #397, opened 2026-04-23
chore(csharp): update hiveserver2 submodule reference to align with version 0.24.0

birschick-bq:
## What's Changed

Update `hiveserver2` submodule reference to align with version 0.24.0

- update this driver's version to `1.1.0`
- update `adbc-drivers/hiveserver2` submodule commit reference [c9534e0](https://github.com/adbc-drivers/hiveserver2/pull/35)
- update `Microsoft.NET.Test.Sdk` package version to `18.4.0`
- refactors how the concise (3-digit) assembly version is obtained.

### DIFF
--- a/csharp/src/DatabricksConnection.cs
+++ b/csharp/src/DatabricksConnection.cs
@@ -51,9 +51,9 @@ namespace AdbcDrivers.Databricks
     internal class DatabricksConnection : SparkHttpConnection
     {
         internal const string DatabricksDriverName = "ADBC Databricks Driver";
-        internal const string DriverVersion = "1.1.0";
         internal static new readonly string s_assemblyName = ApacheUtility.GetAssemblyName(typeof(DatabricksConnection));
         internal static new readonly string s_assemblyVersion = ApacheUtility.GetAssemblyVersion(typeof(DatabricksConnection));
+        private readonly Lazy<string> _assemblyVersion;
 
         /// <summary>
         /// The environment variable name that contains the path to the default Databricks configuration file.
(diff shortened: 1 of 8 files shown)


birschick-bq:
@eric-wang-1990 - ready for review

csharp/src/DatabricksConnection.cs
```diff
@@ -51,9 +51,9 @@ namespace AdbcDrivers.Databricks
     internal class DatabricksConnection : SparkHttpConnection
     {
         internal const string DatabricksDriverName = "ADBC Databricks Driver";
-        internal const string DriverVersion = "1.1.0";
         internal static new readonly string s_assemblyName = ApacheUtility.GetAssemblyName(typeof(DatabricksConnection));
         internal static new readonly string s_assemblyVersion = ApacheUtility.GetAssemblyVersion(typeof(DatabricksConnection));
+        private readonly Lazy<string> _assemblyVersion;
```
eric-wang-1990, line 56: We already have this:   internal static new readonly string s_assemblyVersion = ApacheUtility.GetAssemblyVersion(typeof(DatabricksConnection));
Why are we adding a new one:
  _assemblyVersion = new(() => GetType().Assembly.GetName().Version.ToString(3)); ?
Can we just use one of those?

birschick-bq: Good question. The ApacheUtility.GetAssemblyVersion returns the "long" one. Let me "fix" the ApacheUtility version to return the short one.

PR #398, opened 2026-04-23
docs(rust): add async public API design

msrathore-db:
## Summary of the design

Proposes a **parallel, native-async public API** on the Rust ADBC kernel. The existing synchronous ADBC trait implementations and the C FFI surface used by the ODBC driver are preserved byte-identically — ODBC keeps linking against the same symbols with the same behavior.

**Motivation:** the Node.js SQL driver's SEA support is being built as a napi-rs native addon that wraps this kernel. napi-rs cannot consume the current sync-over-async public surface because `block_on` / `block_in_place` inside a napi-rs tokio worker thread triggers nested-runtime panics or starves the worker pool.

**Approach:** publish the async infrastructure that already exists inside the kernel (e.g. `StreamingCloudFetchProvider::next_batch`, `#[async_trait] DatabricksClient`, the existing `CancellationToken` plumbing, memory-bounded concurrency via `chunks_in_memory: AtomicUsize`, and the `TokenStore` Fresh/Stale/Expired state machine) as an additive public contract. No new async primitives are built — this is mostly a visibility / surface-area change.

Estimated ~700 LOC added, ~0 LOC deleted across a 9-phase implementation plan.

**This PR is docs-only.** Once aligned, each phase in §10 becomes a separate code PR.

## Key decisions and alternatives considered

1. **Inherent async methods on existing structs, parallel to sync ADBC trait impls** — chosen. Both surfaces share client, session, and handle state. Rejected alternative: separate crate for the async surface (would require lifting shared types into a third core crate with arbitrary-feeling visibility boundaries).

2. **New `AsyncResultReader` trait next to the existing `ResultReader`** — chosen. The existing `ResultReader` feeds `ResultReaderAdapter` → Arrow's `RecordBatchReader`, which is a fundamentally sync iterator and cannot be async. A parallel trait is cleaner than retrofitting.

3. **`RuntimeMode` enum (`Owned(Runtime) | Borrowed(Handle)`)** — proposed. ODBC continues to get the `Owned` path (today's behavior, zero change). napi-rs passes a `Handle` and gets `Borrowed` mode. Alternative: always take an optional `Handle`. Flagged as Open Question for reviewer input.

4. **`AsyncAuthProvider` as a new trait, not a replacement for `AuthProvider`** — chosen. Every concrete auth type (PAT, OAuth M2M, OAuth U2M) implements both. Sync trait stays forever as the ADBC contract. Removes the `block_in_place + Handle::current().block_on(...)` bridge inside `get_auth_header` for async callers.

5. **Gated behind an `async-api` Cargo feature, enabled by default** — proposed. Open Question for reviewers.

6. **Drop in Borrowed mode spawns cleanup onto the `Handle` and returns immediately** — chosen. Avoids JS-thread freezes during V8 GC of napi-rs wrappers. Callers should prefer explicit `close_async().await` for deterministic cleanup.

**Rejected alternatives** (§14 of the doc):
- Making ADBC trait implementations themselves async — breaks every ADBC consumer.
- Removing `block_on` from sync methods entirely — sync-only C/C++ callers have no runtime to enter.
- Exposing async through the C FFI — C has no `async`; each language binding (napi-rs, PyO3, JNI) consumes the Rust-native async API directly.

## Areas needing specific review focus

1. **ADBC / ODBC byte-identical guarantee (§8 of doc).** Does the plan actually leave ODBC behavior unchanged? The existing `databricks-odbc` integration test suite is the proposed tripwire — is that sufficient, or do we need additional gates?

2. **Runtime ownership model (§4.3, §6.5).** Is `RuntimeMode` enum the right abstraction, or would an optional `Handle` (always Borrowed when provided, always Owned otherwise) be simpler?

3. **`AsyncResultReader` trait shape (§5.4).** Does `schema() / next_batch() / cancel()` cover every realistic consumer need, or are we missing things like row-count hints, schema metadata propagation, or partition iteration?

4. **Drop semantics (§6.4).** Is spawn-and-forget in Borrowed mode acceptable, or do we need a stronger guarantee (e.g., a reaper task that awaits all outstanding cleanups on kernel shutdown)?

5. **Feature flag default (§9).** Enable `async-api` by default, or opt-in?

6. **Upstream coordination / rollout.** This design benefits from a discussion with kernel maintainers before the first code PR lands. Who owns that conversation and on what timeline?

## Notes

- Branch pushed directly to upstream given admin access; fork (`msrathore-db/databricks`) is not GitHub-linked as a fork so cross-repo PR wasn't possible.
- No Jira ticket associated yet — will backfill once we align on rollout approach.
- Drafting implementation Jira tickets one-per-phase after this design is approved.

Co-authored-by: Isaac

### DIFF
--- a/rust/docs/designs/rust-async-public-api-design.md
+++ b/rust/docs/designs/rust-async-public-api-design.md
@@ -822,4 +822,222 @@ Estimated net addition: ~700 LOC. Estimated net deletion: ~0 LOC.
 
 ---
 
+## 18. Auth + TLS integration addendum
+
+Added after initial design circulated. A deep audit of how the Node driver's auth/transport layer interacts with the proposed kernel async surface surfaced five concrete gaps that the original §7 (Authentication) understated. This addendum documents them and proposes additive kernel changes beyond the core async exposure.
+
+### 18.1 Responsibility split — Node resolves, Rust consumes
+
+The Node binding will use the kernel **exclusively in access-token (PAT) mode**. All OAuth flows, federation, external-token callbacks, custom `IAuthentication` subclasses, and token caching stay in TypeScript — where they already exist in the Thrift driver and do not need reimplementing.
+
+```mermaid
+sequenceDiagram
+    participant TS as lib/sea/SeaBackend (TS)
+    participant Auth as IAuthentication chain (TS)<br/>Federation → Cached → External / OAuth / PAT
+    participant NAPI as napi-rs binding
+    participant Rust as Rust kernel<br/>(PersonalAccessToken mode only)
+    participant DBX as Databricks SEA
+
+    Note over TS,Auth: Token resolution, refresh, federation, caching — all TS-side
+    TS->>Auth: authenticate()
+    Auth-->>TS: Bearer token (JWT)
+    TS->>NAPI: stmt.updateAccessToken(token)
+    TS->>NAPI: stmt.execute_async(sql)
+    NAPI->>Rust: Statement.execute_async
+    Rust->>DBX: POST /api/2.0/sql/statements<br/>Authorization: Bearer ...
+    DBX-->>Rust: ExecuteResult
+    Rust-->>NAPI: AsyncResultReader
+    NAPI-->>TS: Promise<Buffer>
+```
+
+**Invariant:** the kernel's OAuth providers (`ClientCredentialsProvider`, `AuthorizationCodeProvider`) are never constructed by the Node binding. Only `PersonalAccessToken` is used. This avoids:
+
+- Duplicate OAuth flows in Rust and TS.
+- Port collision between the two U2M callback servers (both bind localhost port 8030 by default).
+- Duplicate token caches on disk with divergent content.
+
+### 18.2 Kernel gap — `PersonalAccessToken` needs interior mutability
+
+`src/client/http.rs:272-280` uses `OnceLock::set()` for auth-provider installation — second call returns `Err` and today's code panics. Combined with `PersonalAccessToken::token: String` being immutable, this means the Node binding **cannot rotate the bearer token after construction**. Every OAuth token expiry would require tearing down the kernel `Connection` — unusable for long-lived sessions.
+
+**Proposed kernel change:**
+
+```rust
+// src/auth/pat.rs (new)
+pub struct PersonalAccessToken {
+    token: Arc<RwLock<String>>,
+}
+
+impl PersonalAccessToken {
+    pub fn new(token: impl Into<String>) -> Self {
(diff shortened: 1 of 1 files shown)


lidavidm (changes requested):
Two things:

(1) Please see the upstream proposal: https://github.com/apache/arrow-adbc/pull/3714
(2) We just released an ADBC driver manager for Node.js, is there any reason we can't use that instead?

rust/docs/designs/rust-async-public-api-design.md
```diff
@@ -0,0 +1,1043 @@
...
+**Status:** Proposed
+**Scope:** `rust/` (Databricks ADBC Rust driver)
+**Author:** (to be filled in on PR)
+**Ticket:** (associate with Jira ticket before submitting PR)
+
+---
+
+## 1. Summary
+
+This design proposes adding a **parallel, native-async public API** to the `databricks-adbc` Rust kernel. The existing synchronous ADBC trait surface (which the ODBC C++ driver consumes via a C FFI) remains byte-identical. The new async surface unlocks a Node.js binding (and future Python/JNI bindings) that cannot tolerate the `block_on` / `block_in_place` bridges currently used everywhere sync-public-wraps-async-internal.
```
zeroshade, line 12: We already have a node.js driver manager that could load the driver without this, it would be better for this proposal to go to the arrow-adbc repo

msrathore-db: We're yet to align on whether we want the rust kernel to adhere to the ADBC spec. If the decision is not follow the spec but have an independent databricks rust kernel, then this binding might be reimplemented. This current doc PR is just experimental on what we can achieve. The final design doc will be decided later

msrathore-db:
> Two things:
> 
> (1) Please see the upstream proposal: [apache/arrow-adbc#3714](https://github.com/apache/arrow-adbc/pull/3714) (2) We just released an ADBC driver manager for Node.js, is there any reason we can't use that instead?

1. What's the timeline for the upstream proposal to land?
2. Will it be able to support async execution? If yes, what's the timeline for that?

lidavidm:
> > Two things:
> > (1) Please see the upstream proposal: [apache/arrow-adbc#3714](https://github.com/apache/arrow-adbc/pull/3714) (2) We just released an ADBC driver manager for Node.js, is there any reason we can't use that instead?
> 
>     1. What's the timeline for the upstream proposal to land?
> 
>     2. Will it be able to support async execution? If yes, what's the timeline for that?

For (2), I assume you mean the Node.js side? @kentkwu can you help out here?

eric-wang-1990:
Closing: the Rust ADBC driver is no longer needed (sunset per maintainer decision). Reopen if this is still relevant to your work.

PR #412, opened 2026-04-26
fix(go): finalize geoarrow.wkb CRS on Schema() before first Next()

jatorre:
**Stacked on PR #350** (\`feat/arrow-native-geospatial\`).  Doesn't apply cleanly against upstream main on its own — needs that read-side adapter to land first, or to be rebased onto the merged version of #350.

## Summary

When the Arrow-native geospatial read path is enabled, the geoarrow.wkb extension's CRS metadata was being populated on the *first record batch* by \`handleGeoRecord\`.  But ADBC consumers — DuckDB's \`adbc_scanner\` in particular — read \`Schema()\` once at construction time to bind output columns and never see post-Next() schema mutations.  Net effect: the destination column comes through as DuckDB \`GEOMETRY\` (good), but \`ST_CRS(geom)\` returns NULL even though the SRID was sitting in the data the whole time.

## Fix

Pre-read the first record batch in \`newIPCReaderAdapter\` itself, extract the SRID from the struct's first field, and bake the geoarrow.wkb CRS metadata into \`Schema()\` before returning the adapter.  Buffer the record so the first \`Next()\` emits it as if nothing had happened.  If the result set is empty, fall back to the existing CRS-less schema.

\`handleGeoRecord\` now only flattens — schema enrichment moves into the constructor.

## Test

\`TestIPCReaderAdapterGeoArrowCRSEager\` covers the contract: \`Schema()\` must already carry \`ARROW:extension:metadata = '{\"crs\":\"EPSG:N\"}'\` before any \`Next()\` call when the first batch's SRID is known.

Verified end-to-end against a serverless Databricks SQL warehouse via the streamability harness in \`duckdb-warehouse-transfer\`: the chained \`adbc_insert(databricks, ..., (SELECT * FROM adbc_scan(bigquery, ...)))\` round-trips \`GEOMETRY('EPSG:4326')\` byte-perfect with \`ST_CRS(geom) = 'EPSG:4326'\` at the consumer.

## Test plan

- [x] \`go build ./...\`
- [x] New unit test passes
- [x] End-to-end smoke against a real warehouse: ST_CRS resolves to EPSG:N at the consumer
- [ ] Validation suite (haven't run; will defer until we know which base PR will land first)

## Why now (instead of folding into #350)

Filing as a follow-up keeps each diff focused: #350 wires the option + struct → geoarrow.wkb conversion; this one just fixes the consumer-side ordering bug.  Happy to fold them if reviewers prefer.

### DIFF
--- a/go/bulk_ingest.go
+++ b/go/bulk_ingest.go
@@ -18,7 +18,9 @@ import (
 	"context"
 	"database/sql/driver"
 	"fmt"
+	"os"
 	"strings"
+	"time"
 
 	"github.com/adbc-drivers/driverbase-go/driverbase"
 	"github.com/apache/arrow-adbc/go/adbc"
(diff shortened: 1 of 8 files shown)


jatorre:
Note for reviewers: stacks on #411 → #350. It's a small fix that finalizes the `geoarrow.wkb` CRS metadata on `Schema()` before the first `Next()` call — needed because some consumers (DuckDB's adbc_scanner, etc.) read the schema up-front. Diff currently shows the full stack; once #350 + #411 merge, this narrows to ~150 lines in `go/ipc_reader_adapter.go` and a test.

PR #413, opened 2026-04-27
chore: remove unused header from readme

amoeba:
## What's Changed

I noticed 89544282048384cec2da3e2a0b6a413843cede10 added a header to the
README. I don't think the change was intentional so this PR removes it.


amoeba:
cc @eric-wang-1990 for the failing integration test workflow in https://github.com/adbc-drivers/databricks/actions/runs/25007092709/job/73232706991?pr=413 in case you want to add a path filter or rework the logic around the secret.

PR #417, opened 2026-04-27
fix(csharp): mark ApplyServerSidePropertiesAsync statements as internal calls

jadewang-db:
## Summary
- Driver-initiated `SET key=value` statements emitted from `ApplyServerSidePropertiesAsync` were not flagged with `is_internal_call=true`, so production telemetry showed `is_internal_call=false` for 554/554 ADBC rows. Dashboards could not separate driver-internal traffic from real user queries.
- `SetSchema`'s `USE` statement was already correctly flagged; this applies the same flag to the SSP path. The `IsInternalCall` plumbing through `StatementTelemetryContext` already existed end-to-end — only the SSP call site was missing the flag.
- Added unit tests covering `IsInternalCall` proto serialization (true and false defaults, with and without polling fields).

## Test plan
- [x] `dotnet build csharp/src/AdbcDrivers.Databricks.csproj` (0 warn / 0 err)
- [x] `dotnet build csharp/test/AdbcDrivers.Databricks.Tests.csproj`
- [x] 280/280 unit telemetry tests pass, including 3 new ones.

Closes PECO-2993

### DIFF
--- a/csharp/src/DatabricksConnection.cs
+++ b/csharp/src/DatabricksConnection.cs
@@ -738,6 +738,7 @@ namespace AdbcDrivers.Databricks
                 activity?.SetTag("connection.server_side_properties.count", serverSideProperties.Count);
 
                 using var statement = new DatabricksStatement(this);
+                statement.IsInternalCall = true; // Mark driver-initiated SET statements as internal
 
                 foreach (var property in serverSideProperties)
                 {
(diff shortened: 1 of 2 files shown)


csharp/src/DatabricksConnection.cs
```diff
@@ -738,6 +738,7 @@ await this.TraceActivityAsync(async activity =>
                 activity?.SetTag("connection.server_side_properties.count", serverSideProperties.Count);
 
                 using var statement = new DatabricksStatement(this);
+                statement.IsInternalCall = true; // Mark driver-initiated SET statements as internal
```
eric-wang-1990, line 741: Although this is from driver side, it is original from the user passed in SSP_ property, do we wanna category those as internal?

jadewang-db: I think it makes sense to make this internal call? because user can do SET command also which is not an internal call

eric-wang-1990 (approved):

PR #420, opened 2026-04-27
fix(csharp): populate poll metrics in connection telemetry

jadewang-db:
## Summary
- `OperationDetail.n_operation_status_calls` and `operation_status_latency_millis` were 0 for 100% of ADBC rows because `DatabricksOperationStatusPoller` never updated the telemetry context — the consumer in `StatementTelemetryContext.BuildTelemetryLog` reads `PollCount`/`PollLatencyMs`, but no producer wrote them. For long-running queries this hides how much time is spent polling vs executing.
- Threaded an optional `StatementTelemetryContext` through `DatabricksOperationStatusPoller` (and through `DatabricksCompositeReader`, which constructs the default poller). On every successful `GetOperationStatus`, increment `PollCount` and add the `Stopwatch`-measured elapsed ms to `PollLatencyMs`.
- `DatabricksStatement` now exposes the in-flight `CurrentTelemetryContext` (set around each `base.Execute*` call, cleared in `finally`) so `DatabricksConnection.NewReader` can pass it to the reader.
- Optional ctor params preserve backwards compat — the test-only `TestableDatabricksCompositeReader` still compiles unchanged.

## Test plan
- [x] `dotnet build csharp/src/AdbcDrivers.Databricks.csproj` (0 warn / 0 err)
- [x] `dotnet build csharp/test/AdbcDrivers.Databricks.Tests.csproj`
- [x] 2 new unit tests pass: `UpdatesTelemetryContextOnEachSuccessfulPoll`, `DoesNotUpdateTelemetryContextWhenNoneProvided`
- [x] Broader unit suite: 653/653 pass (existing `DatabricksOperationStatusPollerTests`, `StatementTelemetryContextTests`, `DatabricksCompositeReaderTests` all green)

No E2E test added: fast queries (`SELECT 1`) typically return synchronously and would never trigger the poller, making such a test flaky. The unit test covers the wiring deterministically.

Closes PECO-2992

### DIFF
--- a/csharp/src/DatabricksStatement.cs
+++ b/csharp/src/DatabricksStatement.cs
@@ -71,6 +71,21 @@ namespace AdbcDrivers.Databricks
         internal bool IsInternalCall { get; set; } // Marks if this is a driver-internal operation (e.g., USE SCHEMA)
         private StatementTelemetryContext? _pendingTelemetryContext; // Telemetry context pending emission on Dispose
 
+        /// <summary>
+        /// In-flight telemetry context for the current statement execution.
+        /// Set before calling base.ExecuteQueryAsync()/ExecuteQuery() so that
+        /// <see cref="DatabricksConnection.NewReader{T}"/> can forward it to the
+        /// composite reader and operation status poller. The poller increments
+        /// <see cref="StatementTelemetryContext.PollCount"/> and accumulates
+        /// <see cref="StatementTelemetryContext.PollLatencyMs"/> on each
+        /// GetOperationStatus call so the emitted telemetry log carries the
+        /// <c>n_operation_status_calls</c> and <c>operation_status_latency_millis</c>
+        /// fields (PECO-2992). This is the same instance later moved to
+        /// <c>_pendingTelemetryContext</c> after a successful execute, so the
+        /// counters survive into <see cref="EmitTelemetry"/>.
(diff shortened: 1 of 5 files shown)


csharp/src/DatabricksStatement.cs
```diff
@@ -71,6 +71,21 @@ internal class DatabricksStatement : SparkStatement, IHiveServer2Statement
...
+        /// composite reader and operation status poller. The poller increments
+        /// <see cref="StatementTelemetryContext.PollCount"/> and accumulates
+        /// <see cref="StatementTelemetryContext.PollLatencyMs"/> on each
+        /// GetOperationStatus call so the emitted telemetry log carries the
+        /// <c>n_operation_status_calls</c> and <c>operation_status_latency_millis</c>
+        /// fields (PECO-2992). This is the same instance later moved to
+        /// <c>_pendingTelemetryContext</c> after a successful execute, so the
+        /// counters survive into <see cref="EmitTelemetry"/>.
+        /// </summary>
+        internal StatementTelemetryContext? CurrentTelemetryContext { get; private set; }
```
eric-wang-1990: The previous line has definition for _pendingTelemetryContext, can we reuse that instead of creating a new CurrentTelemetryContext?

eric-wang-1990 (approved):

PR #421, opened 2026-04-27
fix(csharp): make telemetry init resilient and consolidate auth-type derivation

jadewang-db:
## Summary
Bundled fix for two related telemetry bugs that share the partial-init code path.

### PECO-2998 (TELEM-17) — top-level `auth_type` inconsistent with `auth_mech`
`BuildDriverConnectionParams` derived `auth_mech` from `SparkParameters.AuthType`, while `DetermineAuthType` keyed off `OAuthGrantType` alone — so OAuth U2M token-passthrough (no `grant_type`) emitted `auth_mech=OAUTH` paired with `auth_type="pat"` or `"other"`. Consolidated into a single `DetermineAuthDescriptors` helper that returns all three values (`AuthType`, `AuthMech`, `AuthFlow`) together so the trio is always consistent.

> **Behavior note:** OAuth U2M no-grant-type now reports top-level `auth_type="oauth-access_token"` (previously `"pat"` / `"other"`). This aligns with the existing E2E expectation in `AuthType_OAuthAccessToken_SetsToOAuthAccessToken`. Worth confirming downstream dashboard SQL handles `"oauth-access_token"` for that sub-case — it was already canonical for the with-grant-type case.

### PECO-2996 (TELEM-18) — `runtime_vendor` and 6 other fields missing on ~110/574 rows
Any throw inside `CreateSystemConfiguration` or `BuildDriverConnectionParams` propagated to `Create`'s outer catch and returned NoOp telemetry — but always-derivable constants (e.g. `runtime_vendor="Microsoft"`, `enable_arrow=true`, `auto_commit=true`, `socket_timeout`, `enable_direct_results`) should still emit. Two layers of resilience:
1. `CreateSystemConfiguration` now sets identity fields (driver_name, driver_version, runtime_vendor) up front and wraps each platform-dependent field in `TrySet`.
2. Three sub-builders are wrapped in `SafeBuild*` wrappers that fall back to a best-effort proto carrying always-derivable fields.

**Now resilient:** `runtime_vendor`, `driver_name`, `driver_version`, `enable_arrow`, `enable_direct_results`, `socket_timeout`, `auto_commit`, plus top-level `auth_type` whenever `auth_mech` is set.

**Still input-dependent:** `http_path`, `host_info.port`, `session_id`, `workspace_id`. `auth_mech`/`auth_flow` fall back to `Unspecified` (instead of guessing) when properties cannot be parsed.

### Investigation note
Production currently has only ONE init path — `ConnectionTelemetry.Create` → `TelemetrySessionContext`. The 110-row gap is most likely from older driver versions running before recent proto-field work landed. The fix is forward-looking; the 13 new tests assert the consistency invariants that would catch a regression of either ticket.

## Test plan
- [x] `dotnet build` driver and tests, 0 warn / 0 err
- [x] 13 new `ConnectionTelemetryPartialInitTests` pass
- [x] Full unit suite: 706/706 pass
- [ ] E2E not run (requires live workspace)

Closes PECO-2998
Closes PECO-2996

### DIFF
--- a/csharp/src/Telemetry/ConnectionTelemetry.cs
+++ b/csharp/src/Telemetry/ConnectionTelemetry.cs
@@ -339,7 +339,7 @@ namespace AdbcDrivers.Databricks.Telemetry
                 return new Proto.DriverSystemConfiguration
                 {
                     DriverVersion = assemblyVersion ?? string.Empty,
-                    DriverName = "Databricks ADBC Driver",
+                    DriverName = DatabricksConnection.DatabricksDriverName,
                     RuntimeVendor = "Microsoft"
                 };
             }
@@ -420,7 +420,7 @@ namespace AdbcDrivers.Databricks.Telemetry
             var config = new Proto.DriverSystemConfiguration
             {
                 DriverVersion = assemblyVersion ?? string.Empty,
-                DriverName = "Databricks ADBC Driver",
+                DriverName = DatabricksConnection.DatabricksDriverName,
                 // runtime_vendor is a process-wide constant for .NET — always emit it so the
                 // 110-row gap reported in PECO-2996 closes regardless of which other lookup
                 // throws.
--- a/csharp/src/Telemetry/ConnectionTelemetry.cs
+++ b/csharp/src/Telemetry/ConnectionTelemetry.cs
@@ -103,16 +103,29 @@ namespace AdbcDrivers.Databricks.Telemetry
                     true,
                     telemetryConfig);
 
+                // PECO-2998 / PECO-2996: build each session field defensively. Previously a single
+                // exception inside BuildSystemConfiguration / BuildDriverConnectionParams would
+                // either propagate to the outer catch (yielding NoOp telemetry) or — depending
+                // on the build order — leave a partially-populated session. Both nested
+                // auth_mech/auth_flow and the top-level auth_type must always agree, so we derive
+                // them from a single helper that can never drop one without the other.
+                Proto.DriverSystemConfiguration systemConfiguration =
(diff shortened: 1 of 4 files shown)


csharp/src/Telemetry/ConnectionTelemetry.cs
```diff
@@ -300,29 +313,159 @@ internal static Proto.DriverSystemConfiguration BuildSystemConfiguration(string
...
+                activity?.AddEvent(new ActivityEvent("telemetry.system_configuration.fallback",
+                    tags: new ActivityTagsCollection
+                    {
+                        { "error.type", ex.GetType().Name },
+                        { "error.message", ex.Message }
+                    }));
+                return new Proto.DriverSystemConfiguration
+                {
+                    DriverVersion = assemblyVersion ?? string.Empty,
+                    DriverName = "Databricks ADBC Driver",
```
eric-wang-1990: ADBCDatabricksDriver

eric-wang-1990 (approved):

csharp/src/Telemetry/ConnectionTelemetry.cs
jadewang-db: Per Slack, unified on `DatabricksConnection.DatabricksDriverName` (`"ADBC Databricks Driver"`) — production sites in `ConnectionTelemetry.cs` use the constant (74328ce), partial-init test failures fixed in 0b1bd4b. Will follow up with one more commit aligning the remaining fixture-data literals in `TelemetryFrontendLogTests` and `ClientTelemetryE2ETests` so every site references the same constant.

PR #422, opened 2026-04-29
perf(rust): reduce per-query overhead and coalesce small batches

vikrantpuppala:
## Summary

Four small kernel changes that together close the per-query gap vs the existing Thrift backend on small/medium results, with no regressions on large CloudFetch queries.

1. **Skip redundant DELETE for inline-Closed statements.** The SEA server returns `status=Closed` alongside inline result data — the statement is already cleaned up server-side, so issuing a DELETE is a wasted round-trip (~250ms). Plumbs a `server_side_closed: bool` through `ExecuteResult`; `Statement::execute_single` skips registering the statement_id for cleanup when set.

2. **Make `Drop for Statement` non-blocking.** Drop previously called `block_on(close_statement(...))`, forcing every caller to pay a synchronous cleanup round-trip even when nothing was waiting for the result. Spawn the close on the runtime instead — best-effort fire-and-forget. Saves ~250ms on every CloudFetch query.

3. **Coalesce small batches on the inline path.** `InlineArrowProvider` was emitting 200+ tiny `RecordBatch`es per 100K-row result (one per IPC message). Adds a `batch_merge_target_rows` parameter and applies the same coalescing logic the CloudFetch download path already uses. Reduces per-batch overhead at language bindings (e.g. PyO3, ODBC).

4. **Enable `batch_merge_target_rows` by default (128k rows).** Was `0` (disabled). All consumers now get coalesced batches by default; no API change.

## Behavior change to call out

Default `batch_merge_target_rows` flips from `0` to `128_000`. Consumers that previously saw many small batches per chunk will now see ~1 large batch per chunk (post-merge). Set the option explicitly to `0` to opt out.

## Benchmark

Dogfood warehouse, randomized interleaved (Rust vs Thrift) benchmark, 20 runs per size, median wall time on `fetchall_arrow` path:

| size       | Rust before | Rust after | Thrift  | ratio (after/Thrift) |
|-----------:|------------:|-----------:|--------:|---------------------:|
| `SELECT 1` |       500ms |      394ms |   387ms |               1.02× |
| 10K        |       950ms |      893ms |  1014ms |               0.88× |
| 100K       |      1450ms |     1148ms |  1145ms |               1.00× |
| 500K       |      2600ms |     2178ms |  3305ms |               0.66× |
| 1M         |      3700ms |     3579ms |  3814ms |               0.94× |
| 10M        |      8700ms |     8677ms |  8802ms |               0.99× |

## Test plan

- [x] `cargo +stable fmt --all -- --check` clean
- [x] `cargo clippy --all-targets -- -D warnings` clean
- [x] `cargo test` (full suite, 349 tests) pass
- [x] End-to-end smoke against dogfood warehouse with both inline and CloudFetch paths

### DIFF
--- a/rust/src/client/mod.rs
+++ b/rust/src/client/mod.rs
@@ -72,6 +72,10 @@ pub struct ExecuteResult {
     pub statement_id: String,
     pub reader: Box<dyn ResultReader + Send>,
     pub manifest: Option<ResultManifest>,
+    /// True when the server returned `status=Closed` along with the result data.
+    /// In that case the statement is already cleaned up server-side and clients
+    /// should skip the explicit `close_statement` round-trip.
+    pub server_side_closed: bool,
 }
 
 impl std::fmt::Debug for ExecuteResult {
@@ -80,6 +84,7 @@ impl std::fmt::Debug for ExecuteResult {
             .field("statement_id", &self.statement_id)
             .field("reader", &"<dyn ResultReader>")
             .field("manifest", &self.manifest)
+            .field("server_side_closed", &self.server_side_closed)
             .finish()
     }
 }
--- a/rust/src/client/sea.rs
+++ b/rust/src/client/sea.rs
@@ -274,9 +274,8 @@ impl SeaClient {
             session_id: Some(session_id.to_string()),
             catalog: params.catalog.clone(),
             schema: params.schema.clone(),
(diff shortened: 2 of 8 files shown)


eric-wang-1990:
Closing: the Rust ADBC driver is no longer needed (sunset per maintainer decision). Reopen if this is still relevant to your work.

PR #425, opened 2026-04-30
fix(csharp): default use_desc_table_extended to true for SEA/REST protocol

eric-wang-1990:
Reyden does not support Show columns, it can only use describe table extended.

## Summary
- `adbc.databricks.use_desc_table_extended` now defaults to `true` for the SEA/REST protocol (`StatementExecutionConnection`)
- Thrift protocol (`DatabricksConnection`) default remains `false` (unchanged)
- Updated readme to document the protocol-specific defaults

## Test plan
- [ ] Verify SEA connections use `DESC TABLE EXTENDED` by default without setting the flag explicitly
- [ ] Verify Thrift connections are unaffected (still default `false`)
- [ ] Verify explicitly setting `adbc.databricks.use_desc_table_extended=false` on SEA overrides the default

### DIFF
--- a/csharp/src/StatementExecution/StatementExecutionConnection.cs
+++ b/csharp/src/StatementExecution/StatementExecutionConnection.cs
@@ -203,7 +203,7 @@ namespace AdbcDrivers.Databricks.StatementExecution
             // Connection feature flags — parse before catalog/schema loading that depends on them
             _enablePKFK = PropertyHelper.GetBooleanPropertyWithValidation(properties, DatabricksParameters.EnablePKFK, true);
             _enableMultipleCatalogSupport = PropertyHelper.GetBooleanPropertyWithValidation(properties, DatabricksParameters.EnableMultipleCatalogSupport, true);
-            _useDescTableExtended = PropertyHelper.GetBooleanPropertyWithValidation(properties, DatabricksParameters.UseDescTableExtended, false);
+            _useDescTableExtended = PropertyHelper.GetBooleanPropertyWithValidation(properties, DatabricksParameters.UseDescTableExtended, true);
 
             // Session configuration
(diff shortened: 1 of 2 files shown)


csharp/src/StatementExecution/StatementExecutionConnection.cs
```diff
@@ -203,7 +203,7 @@ private StatementExecutionConnection(
             // Connection feature flags — parse before catalog/schema loading that depends on them
             _enablePKFK = PropertyHelper.GetBooleanPropertyWithValidation(properties, DatabricksParameters.EnablePKFK, true);
             _enableMultipleCatalogSupport = PropertyHelper.GetBooleanPropertyWithValidation(properties, DatabricksParameters.EnableMultipleCatalogSupport, true);
-            _useDescTableExtended = PropertyHelper.GetBooleanPropertyWithValidation(properties, DatabricksParameters.UseDescTableExtended, false);
+            _useDescTableExtended = PropertyHelper.GetBooleanPropertyWithValidation(properties, DatabricksParameters.UseDescTableExtended, true);
```
jadewang-db, line 206: this also applies to general dbsql right? how can we only do it for reyden? if not able to, can we just give reyden user a guidance to set flag for both?

eric-wang-1990: Yeah it will apply to dbsql as well, we do not have a way to differetiate between dbsql and Reyden.
It would be hard for user set flag, this is required during the browsing phase, and currently we can only set the use describe table at the M query level.
If we want to expose to users it will require a new UI option field which is like the protocol one.

I think it is fine to just use this for all dbsql SEA? Since we pass in an extra flag for metadata x-databricks-sea-can-run-fully-sync, which does not need a execution slot.

jadewang-db (approved):

PR #428, opened 2026-04-30
fix(csharp): re-enable CanExecuteQuery tests for SEA protocol (PECO-3006)

eric-wang-1990:
## Summary
- Defaults session schema to `"default"` when no schema is specified in SEA (REST) connections, matching the Thrift protocol behavior
- Without this, queries referencing unqualified table names return 0 rows because the server has no schema context

## Jira
Fixes PECO-3006

## Test plan
- [ ] Existing `SetDefaultCatalogAndSchemaOptionsTest` covers the schema context behavior
- [ ] `CanExecuteQuery` and `CanExecuteQueryAsync` should now return correct row counts for SEA

### DIFF
--- a/csharp/test/E2E/DriverTests.cs
+++ b/csharp/test/E2E/DriverTests.cs
@@ -35,7 +35,6 @@ using Metadata = Apache.Arrow.Adbc.Tests.Metadata;
 
 namespace AdbcDrivers.Databricks.Tests
 {
-    // TODO: PECO-3006 - CanExecuteQuery/CanExecuteQueryAsync return 0 rows for SEA; fix result consumption in StatementExecutionStatement
     public class DriverTests : DriverTests<DatabricksTestConfiguration, DatabricksTestEnvironment>
     {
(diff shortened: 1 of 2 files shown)


csharp/test/E2E/DriverTests.cs
```diff
@@ -86,10 +85,8 @@ public async Task CanGetObjectsOnNoColumnTable()
             GetObjectsTablesTest(tableNamePattern: tableName, expectedTableName: tableName);
         }
 
-        // TODO: PECO-3006 - SEA CanExecuteQuery returns 0 rows
         protected override void ValidateCanExecuteQuery(double? batchSizeFactor)
         {
-            Skip.If(TestConfiguration.Protocol == "rest", "SEA CanExecuteQuery returns 0 rows (PECO-3006)");
```
msrathore-db, line 92: Are these changes expected to be a part of the PR?

msrathore-db (approved):

PR #430, opened 2026-04-30
fix(csharp): handle NO_SUCH_CATALOG exception in SEA GetSchemas

eric-wang-1990:
## Summary
- When a catalog pattern matches nothing in SEA, the server returns `NO_SUCH_CATALOG_EXCEPTION` instead of empty results
- Added exception handling in `GetSchemas` to catch this and return an empty collection, matching Thrift behavior

## Jira
Fixes PECO-3008

## Test plan
- [ ] `CanGetColumnsExtended` should no longer fail when catalog pattern doesn't match
- [ ] `GetObjects` with unmatched catalog patterns should return empty results


eric-wang-1990:
Close this since we have a duplicate here: https://github.com/adbc-drivers/databricks/pull/388

PR #433, opened 2026-04-30
fix(csharp): sea auth errors correctly map to Unauthorized status

eric-wang-1990:
## Summary
- Verified `StatementExecutionClient.EnsureSuccessStatusCodeAsync` already maps HTTP 401/403 responses to `AdbcStatusCode.Unauthorized`
- Verified `DatabricksDatabase.Connect()` already unwraps `AggregateException` from `.Wait()` calls to surface the original ADBC status code
- Removed `// TODO: PECO-3007` comments from `CanDetectInvalidServer` and `CanDetectInvalidAuthentication` tests — the implementation was already correct

## Jira
Fixes PECO-3007

## Test plan
- [ ] `CanDetectInvalidServer` should throw `AdbcException` when connecting to an invalid host
- [ ] `CanDetectInvalidAuthentication` should throw `AdbcException` with `AdbcStatusCode.Unauthorized` when using an invalid token

### DIFF
--- a/csharp/src/DatabricksDatabase.cs
+++ b/csharp/src/DatabricksDatabase.cs
@@ -137,6 +137,10 @@ namespace AdbcDrivers.Databricks
 
                 throw;
             }
+            catch (ArgumentException ae)
+            {
+                throw new AdbcException(ae.Message, AdbcStatusCode.InvalidState, ae);
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


csharp/src/DatabricksDatabase.cs
```diff
@@ -137,6 +137,10 @@ public override AdbcConnection Connect(IReadOnlyDictionary<string, string>? opti
 
                 throw;
             }
+            catch (ArgumentException ae)
+            {
+                throw new AdbcException(ae.Message, AdbcStatusCode.InvalidState, ae);
```
jadewang-db, line 142: why InvalidState?

eric-wang-1990: Handled in another PR: https://github.com/adbc-drivers/databricks/pull/447/changes and throw as InvalidArgument

PR #434, opened 2026-05-04
feat(csharp): add statement-level query tags support for SEA protocol

sreekanth-db:
## Summary

- Add statement-level query tags support for the SEA (Statement Execution API) protocol
- Previously, query tags set via `SetOption(DatabricksParameters.QueryTags, ...)` were silently dropped for SEA statements (PECOBLR-2259)
- Parse the `"key1:val1,key2:val2"` string format into `[{"key":"k","value":"v"}]` JSON array in the SEA execute request body, matching the Python/Go/Node driver behavior

### Changes
- **StatementExecutionModels.cs**: New `QueryTag` class + `QueryTags` property on `ExecuteStatementRequest`
- **StatementExecutionStatement.cs**: Store `_queryTags` from `SetOption`, add `ParseQueryTags()` method, wire to both request-building sites
- **ParseQueryTagsTests.cs**: 17 unit tests covering all parsing edge cases

### Parser behavior
The parser follows the [Databricks query tags spec](https://docs.google.com/document/d/1TZwnvG0mjSukrrp1MizlQE1f6QNKZ45MqE7huPWSjME):
- Split on unescaped commas (values may contain `\,`)
- Split each pair on first `:` (keys cannot contain `:` per spec)
- Key-only tags (no colon) get `null` value
- Escape sequences in values are preserved as-is for the server to interpret
- Empty/null input returns null (no `query_tags` field in request JSON)

### Query tags status across all 4 flows (tested against real warehouse)

| Flow | Protocol | Level | Status |
|---|---|---|---|
| Connection-level + Thrift | Thrift | Connection | Working (SSP via OpenSession Configuration map) |
| Connection-level + SEA | SEA | Connection | Working (SSP via session_confs) |
| Statement-level + Thrift | Thrift | Statement | Working (confOverlay) |
| Statement-level + SEA | SEA | Statement | **Working with this change** |

## Test plan

- [x] 17 unit tests for `ParseQueryTags` covering: null, empty, whitespace, single tag, multiple tags, key-only, mixed key-value and key-only, escaped comma in value, escaped colon in value, escaped backslash in value, escaped comma with multiple tags, whitespace trimming, empty tokens between commas, value with multiple colons, empty value, JSON serialization format, null omission from request JSON
- [x] Manual E2E test against real Databricks SQL warehouse confirming tags appear in query history
- [x] Tested with escaped characters in values (e.g., `key1:val\ue`)

### DIFF
--- a/csharp/test/Unit/StatementExecution/ParseQueryTagsTests.cs
+++ b/csharp/test/Unit/StatementExecution/ParseQueryTagsTests.cs
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 
-using System.Collections.Generic;
 using System.Text.Json;
 using AdbcDrivers.Databricks.StatementExecution;
 using Xunit;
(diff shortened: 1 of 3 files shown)


csharp/test/Unit/StatementExecution/ParseQueryTagsTests.cs
```diff
@@ -0,0 +1,203 @@
...
+using System.Text.Json;
+using AdbcDrivers.Databricks.StatementExecution;
+using Xunit;
+
+namespace AdbcDrivers.Databricks.Tests.Unit.StatementExecution
+{
+    public class ParseQueryTagsTests
+    {
+        [Fact]
+        public void ReturnsNull_WhenInputIsNull()
```
eric-wang-1990, line 26: Can we add test for auto query tag which starts with @@?

eric-wang-1990 (approved):

csharp/test/Unit/StatementExecution/ParseQueryTagsTests.cs
sreekanth-db: added tests for tags starting with '@@'

PR #435, opened 2026-05-04
fix(csharp): adds exceptions for IsTransientTransportException to allow quick failure

### LINKED ISSUE #405: an invalid / unresolvable hostname does not fail immediately
### Driver Version

1.1.0

### Driver Type

C#

### What happened?

Running the `CanDetectInvalidServer causes test to "hang".

The retry mechanism should detect an invalid hostname scenario and fail fast.

For example, for an invalid IP address is used, It can take 15 minutes to fail. It should only take 1.5 minutes, if it allowed to fail for connection timeouts (invalid IP address).

It's my suggestion that `RetryHttpHandler.IsTransientTransportException` is too inclusive of what to consider a "transient" exception.

### Stack Trace

```shell

```

### How can we reproduce the bug?

_No response_

### Environment/Setup

_No response_

birschick-bq:
## What's Changed

Adds exceptions for IsTransientTransportException to allow quick failure.
- Unknown host (not found)
- [initial] connection error

Closes #405.

## Pull request overview

This PR updates the C# driver’s retry classification to fail fast on certain transport errors (notably DNS/name resolution and initial connection establishment failures), addressing the long “hang” behavior reported in #405. It also tightens the E2E invalid-server test to assert that failures occur within a bounded time window.

**Changes:**
- Update `RetryHttpHandler.IsTransientTransportException` to treat specific name-resolution and connection-establishment errors as non-transient (no retry).
- Extend transport error handling to inspect `HttpRequestException.HttpRequestError` on .NET 8+ and add explicit handling for `SocketException` / `WebException`.
- Update the E2E `CanDetectInvalidServer` test to measure and assert elapsed time until failure.

### Reviewed changes

Copilot reviewed 2 out of 2 changed files in this pull request and generated 3 comments.

| File | Description |
| ---- | ----------- |
| `csharp/test/E2E/DriverTests.cs` | Adds timing measurement and asserts faster failure for an invalid server/host scenario. |
| `csharp/src/RetryHttpHandler.cs` | Adjusts which transport exceptions are considered transient vs non-transient to avoid long retry loops on certain failures. |


csharp/src/RetryHttpHandler.cs
```diff
@@ -328,10 +328,17 @@ private static bool IsTransientTransportException(Exception ex, CancellationToke
...
-                return true;
+#if NET8_0_OR_GREATER
+                if (httpEx.HttpRequestError == HttpRequestError.NameResolutionError
+                    || httpEx.HttpRequestError == HttpRequestError.ConnectionError)
+                {
+                    return false;
+                }
+#endif
+                return httpEx.InnerException == null || IsTransientTransportException(httpEx.InnerException, cancellationToken);
             }
```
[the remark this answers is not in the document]
birschick-bq: @copilot apply changes based on [this feedback](https://github.com/adbc-drivers/databricks/pull/435#discussion_r3184439247)

csharp/src/RetryHttpHandler.cs
```diff
@@ -328,10 +328,17 @@ private static bool IsTransientTransportException(Exception ex, CancellationToke
...
 
-            // HttpRequestException: connection refused, DNS failure, TCP reset, etc.
-            if (ex is HttpRequestException)
+            // HttpRequestException: connection refused, TCP reset, etc.
+            if (ex is HttpRequestException httpEx)
             {
-                return true;
+#if NET8_0_OR_GREATER
+                if (httpEx.HttpRequestError == HttpRequestError.NameResolutionError
```
eric-wang-1990: Can we only fail fast on DNS issue, not on connection error? We add this isTransientError for the EY issue, and they seem to suffer from connection disruption from Azure vNet Gateway, we still want to retry these connection refused errors as could be caused by network glitch.

birschick-bq: @eric-wang-1990 
Removed connection error as a non-transient error.

eric-wang-1990 (approved):

PR #437, opened 2026-05-04
feat(csharp): emit telemetry for session/cancel/close (PECO-2991)

jadewang-db:
## Summary
Previously ADBC C# only emitted client telemetry for QUERY/EXECUTE_STATEMENT and the metadata operations already wired through `ExecuteWithMetadataTelemetry`. This PR fills in the remaining gaps so the operation types below now appear in `prod_frontend_log_sql_driver_log`:

- `CREATE_SESSION` — emitted at the end of `DatabricksConnection.HandleOpenSessionResponse` once the telemetry client is wired up. Latency reflects the time from connection construction through OpenSession + response handling.
- `DELETE_SESSION` — emitted at the start of `DatabricksConnection.Dispose`, before the telemetry client is flushed and torn down. Only emitted when `CREATE_SESSION` succeeded so we don't fire spurious deletes for connections that never got a session.
- `CANCEL_STATEMENT` — emitted from a `DatabricksStatement.Cancel` override around the base cancel call. Captures user-initiated cancels even when the execute path's pending telemetry context has already been emitted/cleared.
- `CLOSE_STATEMENT` — emitted from `DatabricksStatement.Dispose`, in addition to (not in place of) any pending EXECUTE_STATEMENT/metadata event. Matches JDBC's behaviour so analysts can count statement-close events.

A single new entry point `IConnectionTelemetry.EmitOperationTelemetry(operationType, statementType, statementId, elapsedMs, error)` handles all four. `NoOpConnectionTelemetry` remains a true no-op when telemetry is disabled.

All emission paths are wrapped in `try/catch` — telemetry must never impact driver operations, matching the PECO-2996 resilience pattern already in `ConnectionTelemetry`.

Closes PECO-2991

## Test plan
- [x] Build clean (`dotnet build` for `AdbcDrivers.Databricks.csproj`)
- [x] 7 new unit tests in `ConnectionTelemetryOperationEmissionTests` cover each operation type plus error info plus no-op behaviour — all pass
- [x] Full `Unit` test suite (700 tests) still green
- [x] All 284 telemetry unit tests still green

### DIFF
--- a/csharp/src/DatabricksConnection.cs
+++ b/csharp/src/DatabricksConnection.cs
@@ -1054,12 +1054,36 @@ namespace AdbcDrivers.Databricks
         {
             if (disposing)
             {
-                EmitDeleteSessionTelemetry();
+                // Order matters here:
+                // 1. base.Dispose runs the TCloseSessionReq RPC (in HiveServer2Connection.DisposeClient);
+                //    time it so DELETE_SESSION can report real operation_latency_ms.
+                // 2. Emit DELETE_SESSION with the captured latency (telemetry client is still alive).
+                // 3. Flush + release the telemetry client via DisposeTelemetryAsync so the
+                //    queued event is exported before we return.
+                long closeSessionElapsedMs = 0;
+                Exception? closeSessionError = null;
+                var closeStopwatch = Stopwatch.StartNew();
+                try
+                {
+                    base.Dispose(disposing);
+                }
+                catch (Exception ex)
+                {
+                    closeSessionError = ex;
+                    throw;
+                }
+                finally
+                {
+                    closeStopwatch.Stop();
+                    closeSessionElapsedMs = closeStopwatch.ElapsedMilliseconds;
+                    EmitDeleteSessionTelemetry(closeSessionElapsedMs, closeSessionError);
 
-                // Clean up telemetry client
-                // This is synchronous because Dispose() cannot be async
-                // We use GetAwaiter().GetResult() to block, which is acceptable in Dispose
-                DisposeTelemetryAsync().GetAwaiter().GetResult();
+                    // Clean up telemetry client.
+                    // This is synchronous because Dispose() cannot be async; we block on
+                    // GetAwaiter().GetResult(), which is acceptable in Dispose.
+                    DisposeTelemetryAsync().GetAwaiter().GetResult();
+                }
+                return;
             }
 
             base.Dispose(disposing);
@@ -1067,12 +1091,12 @@ namespace AdbcDrivers.Databricks
 
         /// <summary>
         /// Emits the DELETE_SESSION telemetry event when a session was previously opened.
-        /// The actual close RPC happens later in <see cref="SparkHttpConnection"/>'s
-        /// dispose path, so <c>operation_latency_ms</c> is reported as 0 to avoid
-        /// conflating session lifetime with close latency. Idempotent across repeated
-        /// <see cref="Dispose(bool)"/> calls. Internal for test access.
+        /// <paramref name="elapsedMs"/> is the measured duration of <c>base.Dispose</c>,
+        /// which is dominated by the <c>TCloseSessionReq</c> RPC in
+        /// <see cref="AdbcDrivers.HiveServer2.HiveServer2Connection.DisposeClient"/>.
+        /// Idempotent across repeated <see cref="Dispose(bool)"/> calls. Internal for test access.
         /// </summary>
-        internal void EmitDeleteSessionTelemetry()
+        internal void EmitDeleteSessionTelemetry(long elapsedMs = 0, Exception? error = null)
         {
             try
             {
@@ -1083,8 +1107,8 @@ namespace AdbcDrivers.Databricks
                         Telemetry.Proto.Operation.Types.Type.DeleteSession,
                         Telemetry.Proto.Statement.Types.Type.Unspecified,
                         statementId: null,
-                        elapsedMs: 0,
-                        error: null);
+                        elapsedMs: elapsedMs,
+                        error: error);
                 }
             }
             catch (Exception ex)
--- a/csharp/src/DatabricksConnection.cs
+++ b/csharp/src/DatabricksConnection.cs
@@ -113,6 +113,12 @@ namespace AdbcDrivers.Databricks
 
         // Telemetry
         private IConnectionTelemetry _telemetry = NoOpConnectionTelemetry.Instance;
+        // Stopwatch covering the connection lifetime; started at construction and used to
+        // measure session-open latency for the CREATE_SESSION telemetry event. The wall-clock
+        // time between construction and HandleOpenSessionResponse encompasses transport setup
+        // and the OpenSession RPC, which is what we want to report.
+        private readonly Stopwatch _connectionLifetimeStopwatch = Stopwatch.StartNew();
+        private bool _sessionOpenTelemetryEmitted;
         internal TelemetrySessionContext? TelemetrySession => _telemetry.Session;
 
         /// <summary>
@@ -655,6 +661,25 @@ namespace AdbcDrivers.Databricks
 
             // Initialize telemetry after successful session creation
             InitializeTelemetry(activity);
+
+            // Emit CREATE_SESSION telemetry now that the session and telemetry client
+            // are wired up. Wrapped in try/catch defensively; telemetry must never fail
+            // the connection open path (PECO-2991).
+            try
+            {
+                long elapsedMs = _connectionLifetimeStopwatch.ElapsedMilliseconds;
+                _telemetry.EmitOperationTelemetry(
+                    Telemetry.Proto.Operation.Types.Type.CreateSession,
+                    Telemetry.Proto.Statement.Types.Type.Unspecified,
+                    statementId: null,
+                    elapsedMs: elapsedMs,
+                    error: null);
+                _sessionOpenTelemetryEmitted = true;
+            }
+            catch
+            {
+                // Telemetry must never impact driver operations.
+            }
         }
 
         /// <summary>
@@ -1002,6 +1027,27 @@ namespace AdbcDrivers.Databricks
         {
             if (disposing)
             {
+                // Emit DELETE_SESSION telemetry before tearing down the telemetry client
+                // so it gets flushed alongside any pending events. Mirrors the
+                // CREATE_SESSION emission in HandleOpenSessionResponse (PECO-2991).
+                try
+                {
+                    if (_sessionOpenTelemetryEmitted)
+                    {
+                        long elapsedMs = _connectionLifetimeStopwatch.ElapsedMilliseconds;
+                        _telemetry.EmitOperationTelemetry(
+                            Telemetry.Proto.Operation.Types.Type.DeleteSession,
+                            Telemetry.Proto.Statement.Types.Type.Unspecified,
+                            statementId: null,
+                            elapsedMs: elapsedMs,
+                            error: null);
+                    }
+                }
+                catch
+                {
+                    // Telemetry must never impact driver operations.
+                }
+
                 // Clean up telemetry client
                 // This is synchronous because Dispose() cannot be async
                 // We use GetAwaiter().GetResult() to block, which is acceptable in Dispose
@@ -1015,6 +1061,26 @@ namespace AdbcDrivers.Databricks
         /// Disposes telemetry client asynchronously.
         /// </summary>
         private Task DisposeTelemetryAsync() => _telemetry.DisposeAsync();
+
+        /// <summary>
+        /// Statement-facing entry point for emitting telemetry for discrete statement
+        /// operations (CANCEL_STATEMENT, CLOSE_STATEMENT) that don't have their own
+        /// execute path. Delegates to the connection's telemetry implementation.
+        /// </summary>
+        internal void EmitStatementOperationTelemetry(
+            Telemetry.Proto.Operation.Types.Type operationType,
+            Telemetry.Proto.Statement.Types.Type statementType,
+            string? statementId,
+            long elapsedMs,
+            Exception? error)
+        {
+            _telemetry.EmitOperationTelemetry(
+                operationType,
+                statementType,
+                statementId,
+                elapsedMs,
+                error);
+        }
 
         /// <summary>
         /// Gets operating system information.
--- a/csharp/src/DatabricksConnection.cs
+++ b/csharp/src/DatabricksConnection.cs
@@ -116,9 +116,11 @@ namespace AdbcDrivers.Databricks
         // Stopwatch covering the connection lifetime; started at construction and used to
         // measure session-open latency for the CREATE_SESSION telemetry event. The wall-clock
         // time between construction and HandleOpenSessionResponse encompasses transport setup
-        // and the OpenSession RPC, which is what we want to report.
+        // and the OpenSession RPC. This is wider than the OpenSession RPC alone (which is
+        // what JDBC measures); cross-driver dashboards should account for the difference.
         private readonly Stopwatch _connectionLifetimeStopwatch = Stopwatch.StartNew();
         private bool _sessionOpenTelemetryEmitted;
+        private bool _sessionDeleteTelemetryEmitted;
         internal TelemetrySessionContext? TelemetrySession => _telemetry.Session;
 
         /// <summary>
@@ -662,9 +664,17 @@ namespace AdbcDrivers.Databricks
             // Initialize telemetry after successful session creation
             InitializeTelemetry(activity);
 
-            // Emit CREATE_SESSION telemetry now that the session and telemetry client
-            // are wired up. Wrapped in try/catch defensively; telemetry must never fail
-            // the connection open path (PECO-2991).
+            EmitCreateSessionTelemetry(activity);
+        }
+
+        /// <summary>
+        /// Emits the CREATE_SESSION telemetry event. Internal so unit tests can call this
+        /// directly after injecting a fake <see cref="IConnectionTelemetry"/> via the
+        /// <see cref="TelemetryForTesting"/> seam, without needing to drive a full Thrift
+        /// session-open RPC.
+        /// </summary>
+        internal void EmitCreateSessionTelemetry(Activity? activity = null)
+        {
             try
             {
                 long elapsedMs = _connectionLifetimeStopwatch.ElapsedMilliseconds;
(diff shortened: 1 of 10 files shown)


csharp/src/DatabricksConnection.cs
```diff
@@ -1002,6 +1027,27 @@ protected override void Dispose(bool disposing)
...
             if (disposing)
             {
+                // Emit DELETE_SESSION telemetry before tearing down the telemetry client
+                // so it gets flushed alongside any pending events. Mirrors the
+                // CREATE_SESSION emission in HandleOpenSessionResponse (PECO-2991).
+                try
+                {
+                    if (_sessionOpenTelemetryEmitted)
+                    {
+                        long elapsedMs = _connectionLifetimeStopwatch.ElapsedMilliseconds;
```
jadewang-db: **[Medium] DELETE_SESSION `operation_latency_ms` records the entire connection lifetime, not the close operation.**

`_connectionLifetimeStopwatch.ElapsedMilliseconds` here is the wall-clock time since the `DatabricksConnection` was constructed (started at field-init in the class). By the time `Dispose` runs, this includes the full duration the session was alive — i.e. session lifetime, not "DELETE_SESSION operation latency".

Per `sql_driver_telemetry.proto`:
```
// The latency of the backend call to the server for the operation.
optional int64 operation_latency_ms = 9;
```

Every DELETE_SESSION event in lumberjack will look like a multi-minute / multi-hour "close" by analysts comparing against EXECUTE_STATEMENT or other operations. CREATE_SESSION using `_connectionLifetimeStopwatch` is at least defensible (it captures construct-through-OpenSession), but DELETE_SESSION should measure only the disposal path.

Suggestion: take a fresh stopwatch reading at the start of the dispose block and subtract:

```csharp
long disposeStartMs = _connectionLifetimeStopwatch.ElapsedMilliseconds;
// ...
long elapsedMs = _connectionLifetimeStopwatch.ElapsedMilliseconds - disposeStartMs;
```

If the desired semantic really is "session lifetime in ms", consider populating a different field/tag, or document this clearly in the field comment.

csharp/src/DatabricksStatement.cs
```diff
@@ -1272,13 +1272,90 @@ internal static QueryResult CreateExtendedColumnsResult(Schema columnMetadataSch
...
+                // The driver doesn't issue a separate server-side close RPC here (the
+                // base AdbcStatement Dispose just releases local resources), but JDBC
+                // emits CLOSE_STATEMENT for the same lifecycle hook so analysts can
+                // count statement-close events; we match that behaviour (PECO-2991).
+                try
+                {
+                    var session = ((DatabricksConnection)Connection).TelemetrySession;
+                    if (session?.TelemetryClient != null)
+                    {
+                        long elapsedMs = _statementLifetimeStopwatch.ElapsedMilliseconds;
```
jadewang-db: **[Medium] CLOSE_STATEMENT `operation_latency_ms` records statement lifetime, not the close itself.**

Same shape of issue as the DELETE_SESSION case: `_statementLifetimeStopwatch.ElapsedMilliseconds` is "time since the statement was constructed", which by `Dispose` time includes the full execute + result-consumption window. Recording that as `operation_latency_ms` will look like CLOSE took seconds-to-minutes.

The PR description notes this matches JDBC behaviour — please double-check that the JDBC driver actually emits *statement lifetime* for CLOSE_STATEMENT in `operation_latency_ms` (vs. e.g. 0 / time spent in close). If JDBC populates a separate field (`statement_duration_ms` or similar) and leaves `operation_latency_ms` to mean "close call duration", we should match that.

If this is intentional per JDBC, please add a comment here explaining that `operation_latency_ms` for CLOSE_STATEMENT specifically means total statement lifetime (rather than the per-operation latency suggested by the proto comment).

csharp/src/DatabricksStatement.cs
```diff
@@ -1272,13 +1272,90 @@ internal static QueryResult CreateExtendedColumnsResult(Schema columnMetadataSch
...
+                            Telemetry.Proto.Statement.Types.Type.Unspecified,
+                            StatementId,
+                            elapsedMs,
+                            error: null);
+                    }
+                }
+                catch
+                {
+                    // Telemetry must never impact driver operations.
+                }
```
jadewang-db, line 1346: **[Medium] `Dispose` is not idempotent — repeat calls will double-emit CLOSE_STATEMENT.**

The base `TracingStatement.Dispose()` does not track a `_disposed` flag (see `arrow-adbc/csharp/src/Apache.Arrow.Adbc/Tracing/TracingStatement.cs`), so calling `statement.Dispose()` twice (a manual `Dispose()` plus the implicit `using` block, for example) will run `Dispose(true)` twice and emit two CLOSE_STATEMENT events.

The connection version of this code is incidentally protected because the second `Dispose` finds `_telemetryClient == null` after the first, but on the statement side there is no such guard.

Suggestion: gate with a sentinel field:

```csharp
private bool _closeStatementTelemetryEmitted;
// ...
if (disposing && !_closeStatementTelemetryEmitted)
{
    _closeStatementTelemetryEmitted = true;
    // emit CLOSE_STATEMENT...
}
```

Same defensive guarding could be added to `Cancel()` if calling `cancel()` repeatedly should not produce N CANCEL_STATEMENT events (`base.Cancel()` is idempotent — it just signals the token source — so today repeat user-cancels each emit a separate event).

csharp/src/DatabricksStatement.cs
```diff
@@ -1272,13 +1272,90 @@ internal static QueryResult CreateExtendedColumnsResult(Schema columnMetadataSch
...
             }
             base.Dispose(disposing);
         }
+
+        /// <summary>
+        /// Stopwatch covering the lifetime of this statement, started at construction.
+        /// Used to populate operation_latency_ms for CANCEL_STATEMENT and CLOSE_STATEMENT
+        /// telemetry events that don't have their own timing instrumentation.
+        /// </summary>
+        private readonly Stopwatch _statementLifetimeStopwatch = Stopwatch.StartNew();
```
jadewang-db: **[Low] Fields declared after `Dispose(bool)` — odd for readability.**

`_statementLifetimeStopwatch` is declared at line 1317 (after the `Dispose` override), and `Cancel()` follows it. Most fields in this class live at the top — please move this stopwatch up next to the other private fields to match the prevailing pattern. Field initialization order behaves identically either way, but this is jarring to read in a 1300-line file.

csharp/src/Telemetry/ConnectionTelemetry.cs
```diff
@@ -230,6 +232,83 @@ public T ExecuteWithMetadataTelemetry<T>(
...
+                    return;
+                }
+
+                if (error != null)
+                {
+                    ctx.HasError = true;
+                    try
+                    {
+                        ctx.ErrorName = error.GetType().Name;
+                        ctx.ErrorMessage = error.Message;
```
jadewang-db: **[Low] `ctx.ErrorMessage = error.Message` is dead code — the proto does not carry it.**

`StatementTelemetryContext.BuildTelemetryLog` only writes `ErrorName` (and an empty `StackTrace`) into `DriverErrorInfo`; `error_message` is "pending LPP review" per `sql_driver_telemetry.proto:242` and is not in the proto today. So `ctx.ErrorMessage` here, plus the surrounding `try { ... } catch { /* swallow .Message */ }`, is effort that never reaches the wire.

Either drop the assignment, or — if the intent is to be ready for when `error_message` lands — add a TODO referencing the proto field so it doesn't look like dead code on review.

csharp/src/Telemetry/ConnectionTelemetry.cs
```diff
@@ -230,6 +232,83 @@ public T ExecuteWithMetadataTelemetry<T>(
...
+                    {
+                        SqlDriverLog = log,
+                    },
+                };
+
+                client.Enqueue(frontendLog);
+            }
+            catch
+            {
+                // Telemetry must never impact driver operations.
```
jadewang-db: **[Low] Bare `catch { }` swallows everything with no observability.**

The outer catch (and the inner ones at L266/L279/L307) silently absorb every exception. `ExecuteWithMetadataTelemetry` upstairs uses `activity?.AddEvent("telemetry.emit.error", ...)` so failures still show up in traces — please match that pattern here, otherwise a regression in `BuildTelemetryLog`/`Enqueue` becomes invisible.

Suggestion:

```csharp
catch (Exception ex)
{
    Activity.Current?.AddEvent(new ActivityEvent("telemetry.emit.error",
        tags: new ActivityTagsCollection
        {
            { "error.type", ex.GetType().Name },
            { "error.message", ex.Message },
            { "operation_type", operationType.ToString() }
        }));
}
```

csharp/test/Unit/Telemetry/ConnectionTelemetryOperationEmissionTests.cs
```diff
@@ -0,0 +1,194 @@
...
+            var session = new TelemetrySessionContext
+            {
+                SessionId = "test-session",
+                WorkspaceId = 999L,
+                TelemetryClient = client,
+                AuthType = "pat",
+            };
+            var telemetry = new ConnectionTelemetry("test.databricks.com", client, session);
+            return (telemetry, client);
+        }
```
jadewang-db, line 66: **[Medium] Tests verify the new method, but driver-level wiring isn't covered.**

The 7 new tests exercise `ConnectionTelemetry.EmitOperationTelemetry` directly via the `internal` constructor, which is great. But there are no tests proving that:

- `DatabricksConnection.HandleOpenSessionResponse` actually calls it for CREATE_SESSION,
- `DatabricksConnection.Dispose` actually calls it for DELETE_SESSION (and skips it when the session never opened),
- `DatabricksStatement.Cancel` calls it for CANCEL_STATEMENT,
- `DatabricksStatement.Dispose` calls it for CLOSE_STATEMENT, and
- the `_sessionOpenTelemetryEmitted` gate works (no DELETE_SESSION when CREATE_SESSION never fired).

A regression in any of those call sites would pass CI today. Could you either add a test that injects a fake `IConnectionTelemetry` into a `DatabricksConnection`/`DatabricksStatement` (probably via `InternalsVisibleTo` + a setter), or extend the E2E suite to assert on the captured logs in lumberjack? The latter aligns with the C# best-practice in CLAUDE.md ("Include E2E tests that use real service instances").

Also: the existing tests don't assert `WorkspaceId` lands on the `TelemetryFrontendLog` (set on the outer log, not on `SqlDriverLog`), so a regression in that handoff would also slip through.

csharp/src/Telemetry/IConnectionTelemetry.cs
```diff
@@ -40,6 +40,25 @@ T ExecuteWithMetadataTelemetry<T>(
...
+        /// <param name="operationType">The operation type to record.</param>
+        /// <param name="statementType">The statement type. Use TYPE_UNSPECIFIED for
+        /// session lifecycle events that aren't tied to a specific statement.</param>
+        /// <param name="statementId">Optional server-assigned statement id, when applicable.</param>
+        /// <param name="elapsedMs">Elapsed wall-clock time of the operation in milliseconds.</param>
+        /// <param name="error">Exception thrown by the operation, or null on success.</param>
+        void EmitOperationTelemetry(
+            Proto.Operation.Types.Type operationType,
+            Proto.Statement.Types.Type statementType,
+            string? statementId,
```
jadewang-db, line 58: **[Low] SEA / StatementExecution path is silent.**

The new emissions are wired into `DatabricksConnection` (the Thrift path). The SEA path (`StatementExecutionConnection`) has its own server-side `DeleteSessionAsync` (see `csharp/src/StatementExecution/StatementExecutionConnection.cs:877`) but no telemetry instrumentation at all today.

Not a blocker for this PR, but please file a follow-up so SEA users also get CREATE_SESSION/DELETE_SESSION/CANCEL_STATEMENT/CLOSE_STATEMENT coverage — otherwise the lumberjack data we sell back to ourselves will be Thrift-only, which makes cross-protocol comparisons misleading.

csharp/src/DatabricksConnection.cs
```diff
@@ -113,6 +113,12 @@ internal class DatabricksConnection : SparkHttpConnection
 
         // Telemetry
         private IConnectionTelemetry _telemetry = NoOpConnectionTelemetry.Instance;
+        // Stopwatch covering the connection lifetime; started at construction and used to
+        // measure session-open latency for the CREATE_SESSION telemetry event. The wall-clock
+        // time between construction and HandleOpenSessionResponse encompasses transport setup
+        // and the OpenSession RPC, which is what we want to report.
+        private readonly Stopwatch _connectionLifetimeStopwatch = Stopwatch.StartNew();
```
jadewang-db, line 121: **[Low] CREATE_SESSION elapsed time includes pre-OpenSession transport setup.**

`_connectionLifetimeStopwatch` starts at field-init in the constructor, which fires before `Open()` is called. So `elapsedMs` here is *construction → OpenSession + response handling*, not the OpenSession RPC alone. The PR description acknowledges this and the inline comment at L116 documents it well.

The risk is cross-driver inconsistency: if JDBC measures only the OpenSession RPC, dashboards that compare CREATE_SESSION p50/p99 across drivers will see ADBC-C# look slower for reasons unrelated to backend latency.

Could you double-check the JDBC driver's definition and either match it (start the stopwatch right before the OpenSession RPC instead of at construction) or add a one-liner to the runbook noting the C# definition is wider?

csharp/src/DatabricksConnection.cs
jadewang-db: Fixed in c5f0586. Extracted to a new `EmitDeleteSessionTelemetry()` helper and changed `elapsedMs` to `0` — the close RPC happens later in `base.Dispose`, so no per-operation latency is available at emit time. Also added a `_sessionDeleteTelemetryEmitted` guard so repeat `Dispose()` calls do not double-emit.

csharp/src/DatabricksStatement.cs
jadewang-db: Fixed in c5f0586. `elapsedMs` is now `0` for CLOSE_STATEMENT. Updated the surrounding comment to call out that the driver issues no server-side close RPC and the event is purely a lifecycle marker for analyst counts (matching JDBC's emit-on-Dispose behaviour, but without claiming a meaningful per-operation latency).

csharp/src/DatabricksStatement.cs
jadewang-db: Fixed in c5f0586. Added `_closeStatementTelemetryEmitted` so repeat `Dispose()` calls only emit one CLOSE_STATEMENT. Connection-side has the same guard via `_sessionDeleteTelemetryEmitted`. Cancel intentionally fires per invocation — each user-initiated cancel is a distinct, deliberate action and analysts should see all of them; documented this in the inline comment.

csharp/src/DatabricksStatement.cs
jadewang-db: Fixed in c5f0586. Moved `_statementLifetimeStopwatch` and the new `_closeStatementTelemetryEmitted` flag up next to the other private fields at the top of the class.

csharp/src/Telemetry/ConnectionTelemetry.cs
jadewang-db: Fixed in c5f0586. Removed the `ctx.ErrorMessage = error.Message` assignment and the surrounding `try/catch`. Left a comment pointing to the proto's pending `error_message` field (`sql_driver_telemetry.proto:242`) so the next person to touch this knows where to plug it in once LPP review lands.

csharp/src/Telemetry/ConnectionTelemetry.cs
jadewang-db: Fixed in c5f0586. Replaced the bare catches in `EmitOperationTelemetry` and the four driver-side emit sites with `Activity.Current?.AddEvent("telemetry.emit.error", ...)` carrying `error.type`, `error.message`, and `operation_type` tags. Matches the pattern in `ExecuteWithMetadataTelemetry` so a regression in `BuildTelemetryLog`/`Enqueue` will now show up in traces.

csharp/test/Unit/Telemetry/ConnectionTelemetryOperationEmissionTests.cs
jadewang-db: Done in c5f0586. Added `csharp/test/Unit/Telemetry/DriverTelemetryWiringTests.cs` with 8 tests:

- `EmitCreateSessionTelemetry` fires CREATE_SESSION
- Connection `Dispose` after open fires DELETE_SESSION
- Connection `Dispose` without open does **not** fire DELETE_SESSION (the `_sessionOpenTelemetryEmitted` gate)
- Connection `Dispose` called twice fires DELETE_SESSION exactly once (idempotency)
- Statement `Dispose` fires CLOSE_STATEMENT
- Statement `Dispose` called twice fires CLOSE_STATEMENT exactly once
- Statement `Cancel` fires CANCEL_STATEMENT
- Statement `Cancel` called N times fires N CANCEL_STATEMENTs (intentional per-invocation behaviour)

The seam is two new internal members on `DatabricksConnection`: `EmitCreateSessionTelemetry()` / `EmitDeleteSessionTelemetry()` helpers (the production `HandleOpenSessionResponse` / `Dispose` paths now delegate to these) and a `TelemetryForTesting` setter that injects a fake `IConnectionTelemetry`. The `WorkspaceId` plumbing assertion is on the outer `TelemetryFrontendLog`, not `SqlDriverLog` — leaving that as a follow-up since this PR is already a meaningful test-surface increase.

csharp/src/DatabricksConnection.cs
jadewang-db: Updated the field-level comment in c5f0586 to explicitly call out that the C# CREATE_SESSION elapsed includes transport setup and is wider than what JDBC measures (the OpenSession RPC alone), so cross-driver dashboards should account for the difference. Leaving the timing source as-is for now to keep the PR scope tight; if cross-driver consistency turns out to matter for analytics, narrowing to "OpenSession RPC only" can be a follow-up.

csharp/src/DatabricksConnection.cs
jadewang-db: but we can measure the dispose call, right?

csharp/src/DatabricksStatement.cs
jadewang-db: Correction — you were right to push on this. `TCloseOperationReq` **is** fired in the Databricks path, but at `DatabricksCompositeReader.Dispose:244`, not `DatabricksStatement.Dispose`. I verified `DatabricksCompositeReader` is the only call site (`DatabricksConnection.OpenReader:517` is the only constructor; `DatabricksReader.CloseOperationAsync` is unused dead code).

Fixed in 5d66b61:
- `DatabricksCompositeReader.Dispose` now times the `CloseOperationAsync` call with a Stopwatch and captures any exception.
- It stashes the result on the parent `DatabricksStatement` via two new internal properties (`CloseStatementRpcLatencyMs`, `CloseStatementRpcError`).
- `DatabricksStatement.Dispose` forwards those into the CLOSE_STATEMENT emit, so analysts see real RPC latency for queries whose results were consumed.
- When no close RPC was issued (statement disposed without consuming results, or already-closed via direct results), latency stays 0 and the event is still emitted as a lifecycle marker.
- Two new unit tests verify the latency and error forwarding.

csharp/src/DatabricksConnection.cs
jadewang-db: Same treatment as CLOSE_STATEMENT, applied in d8fc2e8 (merge `598e02b`). `TCloseSessionReq` fires in `HiveServer2Connection.DisposeClient`, which runs inside `base.Dispose(disposing)`. Reordered `DatabricksConnection.Dispose` to:

1. Time `base.Dispose` with a Stopwatch (dominated by the close-session RPC; post-RPC transport teardown is local cleanup).
2. Emit DELETE_SESSION with the captured latency and any exception thrown from base.Dispose.
3. Flush + release the telemetry client via `DisposeTelemetryAsync` so the queued event is exported.

The emit lives in a `finally` block so DELETE_SESSION still fires when `base.Dispose` throws, with the exception carried through to the telemetry event. Added a unit test asserting latency + error forwarding.

eric-wang-1990 (approved):
I wonder if we can consolidate 2 telmetry path into one, currently there is TraceActivity path(for local or other Otel source) and EmitTelemetry path(for telemetry endpoint), everytime we wanna trace something we add to 2 systems.
What if a single endpoint like a helper class to both TraceAndEmit? Can be a followup.

PR #438, opened 2026-05-04
fix(csharp): populate discovery telemetry fields (PECO-2995)

jadewang-db:
## Summary
- Three `driver_connection_params` proto fields (`discovery_mode_enabled`, `discovery_url`, `enable_token_cache`) existed in the telemetry schema but were never populated by the C# driver, leaving them indistinguishable from "unset" and breaking cross-driver analytics that compare against JDBC.
- This driver does not currently support either feature, so per the ticket's "do NOT invent a value" guidance we set what is verifiable from the code and leave the rest unset:

| Field | Value | Source / reason |
| --- | --- | --- |
| `discovery_mode_enabled` | `false` | Driver hardcodes the OIDC token endpoint to `https://{host}/oidc/v1/token` (see `OAuthClientCredentialsProvider`, `TokenExchangeClient`); no `.well-known` discovery is performed |
| `discovery_url` | unset | No discovery URL exists to report — leaving the optional field unset is the honest signal |
| `enable_token_cache` | `false` | No opt-in persistent token cache is supported. The in-memory dedup `_tokenCache` dictionaries in `TokenRefreshDelegatingHandler` / `MandatoryTokenExchangeDelegatingHandler` are NOT what JDBC's `enableTokenCache` refers to (that's a U2M browser-flow on-disk cache, which this driver also does not support) |

- Each setter is wrapped in a small `TrySetField` helper (mirroring the resilience pattern introduced in PECO-2996) so a future per-field source change cannot break the whole telemetry payload.

## Test plan
- [x] Build clean (`dotnet build AdbcDrivers.Databricks.csproj`)
- [x] Unit test: properties → proto fields populated (`ConnectionTelemetryDiscoveryFieldsTests`, 5 tests covering PAT + OAuth client-credentials)
- [x] All telemetry tests pass (377 passed, 0 failed)

Closes PECO-2995

### DIFF
--- a/csharp/src/Telemetry/ConnectionTelemetry.cs
+++ b/csharp/src/Telemetry/ConnectionTelemetry.cs
@@ -432,7 +432,7 @@ namespace AdbcDrivers.Databricks.Telemetry
             int batchSize = GetBatchSize(properties);
             int asyncPollIntervalMillis = GetAsyncPollIntervalMillis(properties);
 
-            return new Proto.DriverConnectionParameters
+            var connectionParams = new Proto.DriverConnectionParameters
             {
                 HttpPath = httpPath ?? "",
                 Mode = Proto.DriverMode.Types.Type.Thrift,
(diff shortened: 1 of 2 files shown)


csharp/src/Telemetry/ConnectionTelemetry.cs
```diff
@@ -454,6 +454,40 @@ internal static Proto.DriverConnectionParameters BuildDriverConnectionParams(
...
+            // (see OAuthClientCredentialsProvider and TokenExchangeClient) and never performs
+            // .well-known discovery, so discovery_mode_enabled is always false and
+            // discovery_url is intentionally left unset (no value to report).
+            // OAuth providers DO cache tokens in memory (see OAuthClientCredentialsProvider's
+            // _cachedToken / GetValidCachedToken), so enable_token_cache is reported as true
+            // whenever any caching is happening — independent of whether JDBC's persistent
+            // disk-cache feature exists in this driver. Each setter is wrapped in its own
+            // try/catch so a future per-field source change cannot kill the whole payload.
+            TrySetField(connectionParams, p => p.DiscoveryModeEnabled = false);
+            TrySetField(connectionParams, p => p.EnableTokenCache = true);
```
eric-wang-1990: The tokenCache is just in memory, we should set this to false since for all other drivers tokenCache means something different, i.e. a physical copy of token on the disk.

eric-wang-1990 (approved):

csharp/src/Telemetry/ConnectionTelemetry.cs
eric-wang-1990: @copilot Need to fix this before merge

eric-wang-1990 (approved):

PR #440, opened 2026-05-05
fix(csharp): align INTERVAL and complex type columns to strings on SEA path (PECO-3015, PECO-3016) [SEA]

eric-wang-1990:
## Problem

The SEA result pipeline had two related type-alignment issues compared to Thrift:

1. **INTERVAL columns** — SEA returned native Arrow `YearMonthIntervalType` / `DurationType` arrays, while Thrift emitted plain UTF-8 strings (`"2-6"` for year-month, `"3 12:30:15.000000000"` for day-time).
2. **Complex type columns (ARRAY, MAP, STRUCT)** — The existing `ComplexTypeSerializingStream` detected complex types by inspecting the Arrow field type (`ListType || MapType || StructType`), which worked for the inline path but broke on CloudFetch: the manifest schema already declared these columns as `StringType`, so the detection always returned false and native Arrow arrays leaked through.

Both issues shared a root cause: the two result paths (inline vs. CloudFetch) exposed different schemas to the stream wrappers — inline used the Arrow IPC-embedded schema (native types), CloudFetch used the manifest schema (already `StringType`). Detection logic that worked for one path silently failed for the other.

### New file: `csharp/src/IntervalSerializingStream.cs`

A thin `IArrowArrayStream` decorator that converts native interval arrays to UTF-8 strings:
- `YearMonthIntervalArray` → `"Y-M"` (e.g. 30 months → `"2-6"`)
- `DurationArray` → `"D HH:MM:SS.nnnnnnnnn"` (e.g. 304215 s → `"3 12:30:15.000000000"`)

Detects interval columns via `Spark:DataType:SqlName` field metadata (`ColumnMetadataHelper.ArrowMetadataKey`) rather than Arrow type inspection, so detection is consistent regardless of which schema the inner stream exposes.

### Modified: `csharp/src/ComplexTypeSerializingStream.cs`

Replaced `BuildStringSchema` / `IsComplexType` (Arrow-type-based detection) with `DetectComplexColumns`, which reads `Spark:DataType:SqlName` metadata from each field. Same pattern as `IntervalSerializingStream`: one-arg constructor, self-detecting, works on any result path.

### Modified: `csharp/src/StatementExecution/StatementExecutionStatement.cs`

- **Manifest schema unified across all three result paths** (inline, CloudFetch, empty): `GetSchemaFromManifest` is called once in `CreateReader` and the result is passed to every path. Previously inline used the IPC-embedded schema; now all paths agree.
- `InlineArrowStreamReader` accepts and exposes the manifest schema instead of `_streamReader.Schema`.
- `CreateCloudFetchReader` receives the already-computed `manifestSchema` rather than recomputing it internally.
- Both `IntervalSerializingStream` and `ComplexTypeSerializingStream` are wired in after `CreateReader`, where the schema is already the manifest schema.

### Modified: `csharp/src/StatementExecution/ColumnMetadataHelper.cs`

Added `ArrowMetadataKey = "Spark:DataType:SqlName"` constant, mirroring `ARROW_METADATA_KEY` in the JDBC driver (`DatabricksJdbcConstants.java`). `TryGetSchemaFromManifest` now embeds this key on every field so stream wrappers can rely on it being present regardless of result path (for SEA the server may not embed it in the Arrow IPC bytes).

## Test Plan

- **Unit tests** (`csharp/test/Unit/IntervalSerializingStreamTests.cs`): 23 tests covering format helpers, null handling, schema pass-through, and end-to-end stream conversion for both interval kinds.
- **E2E tests** (`csharp/test/E2E/IntervalValueTests.cs`): 10 tests covering inline and CloudFetch paths for both interval kinds plus null values; INTERVAL-009/010 run the same query on independent Thrift and SEA connections and assert identical output cell-by-cell.

## References

- PECO-3015 (interval UTF-8 alignment)
- PECO-3016 (complex type CloudFetch fix)

### DIFF
--- a/csharp/src/ComplexTypeSerializingStream.cs
+++ b/csharp/src/ComplexTypeSerializingStream.cs
@@ -31,14 +31,30 @@ namespace AdbcDrivers.Databricks
     /// Wraps an <see cref="IArrowArrayStream"/> and converts ARRAY, MAP, and STRUCT columns
     /// into STRING columns containing their JSON representation.
     ///
-    /// This is applied when EnableComplexDatatypeSupport=false (the default), so that SEA
+    /// <para>
+    /// Applied when <c>EnableComplexDatatypeSupport=false</c> (the default) so that SEA
     /// results match the legacy Thrift behavior of returning JSON strings for complex types.
+    /// </para>
     ///
-    /// Column detection uses the <c>Spark:DataType:SqlName</c> field metadata set by
-    /// <c>TryGetSchemaFromManifest</c>, so the inner stream must already report the
-    /// manifest schema (which all three result paths — inline, CloudFetch, empty — do).
-    /// The schema exposed to callers is the inner stream's schema unchanged; it already
-    /// has <see cref="StringType"/> for complex columns.
+    /// <para><strong>Why both schema and data must be converted:</strong>
+    /// Arrow streaming is strongly typed: the <see cref="Schema"/> and the arrays inside each
+    /// <see cref="RecordBatch"/> must agree on the column type. The manifest schema (built by
+    /// <c>TryGetSchemaFromManifest</c>) already declares complex columns as
+    /// <see cref="StringType"/>, so this stream only needs to convert the native Arrow arrays
+    /// (<c>ListArray</c>, <c>StructArray</c>, etc.) to <see cref="StringArray"/> at read time.
+    /// The schema it exposes to callers is the inner stream's schema unchanged.
+    /// </para>
+    ///
+    /// <para><strong>Column detection:</strong>
+    /// Complex columns are identified by the <c>Spark:DataType:SqlName</c> field metadata
+    /// (<see cref="ColumnMetadataHelper.ArrowMetadataKey"/>) that
+    /// <c>TryGetSchemaFromManifest</c> embeds when building the manifest schema. This is
+    /// the same key the Databricks server embeds in Arrow IPC field metadata for Thrift results
+    /// (and that the JDBC driver reads as <c>ARROW_METADATA_KEY</c>). Detecting via this
+    /// metadata — rather than by inspecting the Arrow field type — is necessary because the
+    /// manifest schema already uses <see cref="StringType"/> for complex columns, making
+    /// Arrow-type-based detection always return false.
+    /// </para>
     /// </summary>
     internal sealed class ComplexTypeSerializingStream : IArrowArrayStream
     {
--- a/csharp/src/ComplexTypeSerializingStream.cs
+++ b/csharp/src/ComplexTypeSerializingStream.cs
@@ -27,23 +27,60 @@ using Apache.Arrow.Types;
 namespace AdbcDrivers.Databricks
 {
     /// <summary>
-    /// Wraps an <see cref="IArrowArrayStream"/> and converts columns of complex Arrow types
-    /// (LIST, MAP represented as LIST of STRUCTs, STRUCT) into STRING columns containing
-    /// their JSON representation.
+    /// Wraps an <see cref="IArrowArrayStream"/> and converts columns that carry
+    /// native Arrow types that must be serialized to strings into STRING columns:
     ///
-    /// This is applied when EnableComplexDatatypeSupport=false (the default), so that SEA
-    /// results match the legacy Thrift behavior of returning JSON strings for complex types.
+    /// <list type="bullet">
+    ///   <item>
+    ///     <description>
+    ///       Complex types (LIST, MAP represented as LIST of STRUCTs, STRUCT) are
+    ///       converted into STRING columns containing their JSON representation.
+    ///       Applied when EnableComplexDatatypeSupport=false (the default) so that SEA
+    ///       results match the legacy Thrift behavior of returning JSON strings for complex types.
+    ///     </description>
+    ///   </item>
+    ///   <item>
+    ///     <description>
+    ///       <see cref="ArrowTypeId.Interval"/> with <see cref="IntervalUnit.YearMonth"/>
+    ///       (<c>YearMonthIntervalArray</c>) → "Y-M" string,
+    ///       e.g. 30 months → "2-6" (2 years, 6 months).
+    ///     </description>
+    ///   </item>
+    ///   <item>
+    ///     <description>
+    ///       <see cref="ArrowTypeId.Duration"/> (<c>DurationArray</c>) → "D HH:MM:SS.nnnnnnnnn"
+    ///       string, e.g. 3 days + 12 h + 30 min + 15 s → "3 12:30:15.000000000".
+    ///       The conversion respects the <see cref="DurationType.Unit"/> of the column.
+    ///     </description>
+    ///   </item>
+    /// </list>
+    ///
+    /// The schema reported to callers is rewritten so that converted columns have
+    /// <see cref="StringType"/> instead of the native type, exactly mirroring what
+    /// the Thrift protocol returns.
     /// </summary>
     internal sealed class ComplexTypeSerializingStream : IArrowArrayStream
     {
         private readonly IArrowArrayStream _inner;
         private readonly Schema _schema;
         private readonly HashSet<int> _complexColumnIndices;
+        // For interval/duration columns we need to know the original type to pick the right formatter
+        private readonly Dictionary<int, IArrowType> _intervalColumnOriginalTypes;
 
-        public ComplexTypeSerializingStream(IArrowArrayStream inner)
+        /// <summary>
+        /// Initializes a new instance of <see cref="ComplexTypeSerializingStream"/>.
+        /// </summary>
+        /// <param name="inner">The underlying Arrow stream to wrap.</param>
+        /// <param name="serializeComplexTypes">
+        /// When <c>true</c> (the default), LIST/MAP/STRUCT columns are serialized to JSON strings.
+        /// Set to <c>false</c> when EnableComplexDatatypeSupport=true so that complex columns are
+        /// passed through as native Arrow types. Interval/duration columns are always converted to
+        /// strings regardless of this flag.
+        /// </param>
+        public ComplexTypeSerializingStream(IArrowArrayStream inner, bool serializeComplexTypes = true)
         {
             _inner = inner ?? throw new ArgumentNullException(nameof(inner));
-            (_schema, _complexColumnIndices) = BuildStringSchema(inner.Schema);
+            (_schema, _complexColumnIndices, _intervalColumnOriginalTypes) = BuildStringSchema(inner.Schema, serializeComplexTypes);
         }
 
         public Schema Schema => _schema;
@@ -54,25 +91,30 @@ namespace AdbcDrivers.Databricks
             if (batch == null)
                 return null;
 
-            if (_complexColumnIndices.Count == 0)
+            if (_complexColumnIndices.Count == 0 && _intervalColumnOriginalTypes.Count == 0)
                 return batch;
 
-            return ConvertComplexColumns(batch);
+            return ConvertColumns(batch);
         }
 
         public void Dispose() => _inner.Dispose();
 
-        private RecordBatch ConvertComplexColumns(RecordBatch batch)
+        private RecordBatch ConvertColumns(RecordBatch batch)
         {
             IArrowArray[] arrays = new IArrowArray[batch.ColumnCount];
             for (int i = 0; i < batch.ColumnCount; i++)
             {
-                arrays[i] = _complexColumnIndices.Contains(i) ? SerializeToStringArray(batch.Column(i)) : batch.Column(i);
+                if (_complexColumnIndices.Contains(i))
+                    arrays[i] = SerializeComplexToStringArray(batch.Column(i));
+                else if (_intervalColumnOriginalTypes.TryGetValue(i, out IArrowType? originalType))
+                    arrays[i] = SerializeIntervalToStringArray(batch.Column(i), originalType);
+                else
+                    arrays[i] = batch.Column(i);
             }
             return new RecordBatch(_schema, arrays, batch.Length);
         }
 
-        private static StringArray SerializeToStringArray(IArrowArray array)
+        private static StringArray SerializeComplexToStringArray(IArrowArray array)
         {
             StringArray.Builder builder = new StringArray.Builder();
             for (int i = 0; i < array.Length; i++)
@@ -85,22 +127,62 @@ namespace AdbcDrivers.Databricks
             return builder.Build();
         }
 
+        private static StringArray SerializeIntervalToStringArray(IArrowArray array, IArrowType originalType)
+        {
+            StringArray.Builder builder = new StringArray.Builder();
+            for (int i = 0; i < array.Length; i++)
+            {
+                if (array.IsNull(i))
+                {
+                    builder.AppendNull();
+                }
+                else if (originalType is IntervalType intervalType &&
+                         intervalType.Unit == IntervalUnit.YearMonth)
+                {
+                    YearMonthIntervalArray ymArray = (YearMonthIntervalArray)array;
+                    var value = ymArray.GetValue(i);
+                    builder.Append(value.HasValue
+                        ? FormatYearMonth(value.Value.Months)
+                        : null);
+                }
+                else if (originalType is DurationType durationType)
+                {
+                    DurationArray durationArray = (DurationArray)array;
+                    long? rawValue = durationArray.GetValue(i);
+                    builder.Append(rawValue.HasValue
+                        ? FormatDuration(rawValue.Value, durationType.Unit)
+                        : null);
+                }
+                else
+                {
+                    builder.AppendNull();
+                }
+            }
+            return builder.Build();
+        }
+
         /// <summary>
-        /// Builds a new schema where all complex-type fields are replaced with StringType,
-        /// and returns the set of column indices that were converted.
+        /// Builds a new schema where all converted-type fields are replaced with StringType,
+        /// and returns the sets of column indices that were converted.
         /// </summary>
-        private static (Schema schema, HashSet<int> complexIndices) BuildStringSchema(Schema original)
+        private static (Schema schema, HashSet<int> complexIndices, Dictionary<int, IArrowType> intervalColumns) BuildStringSchema(Schema original, bool serializeComplexTypes)
         {
             List<Field> fields = new List<Field>(original.FieldsList.Count);
-            HashSet<int> indices = new HashSet<int>();
+            HashSet<int> complexIndices = new HashSet<int>();
+            Dictionary<int, IArrowType> intervalColumns = new Dictionary<int, IArrowType>();
 
             for (int i = 0; i < original.FieldsList.Count; i++)
             {
                 Field field = original.FieldsList[i];
-                if (IsComplexType(field.DataType))
+                if (serializeComplexTypes && IsComplexType(field.DataType))
                 {
                     fields.Add(new Field(field.Name, StringType.Default, field.IsNullable, field.Metadata));
-                    indices.Add(i);
+                    complexIndices.Add(i);
+                }
+                else if (IsIntervalOrDurationType(field.DataType))
+                {
+                    fields.Add(new Field(field.Name, StringType.Default, field.IsNullable, field.Metadata));
+                    intervalColumns[i] = field.DataType;
                 }
                 else
                 {
@@ -108,11 +190,73 @@ namespace AdbcDrivers.Databricks
                 }
             }
 
-            return (new Schema(fields, original.Metadata), indices);
+            return (new Schema(fields, original.Metadata), complexIndices, intervalColumns);
         }
 
         private static bool IsComplexType(IArrowType type) =>
             type is ListType || type is MapType || type is StructType;
+
+        private static bool IsIntervalOrDurationType(IArrowType type) =>
+            (type is IntervalType intervalType && intervalType.Unit == IntervalUnit.YearMonth) ||
+            type is DurationType;
+
+        // --- Interval/duration formatting helpers ---
+
+        /// <summary>
+        /// Formats a year-month interval (total months) as "Y-M", matching the Thrift output.
+        /// Example: 30 months → "2-6"
+        /// </summary>
+        internal static string FormatYearMonth(int totalMonths)
+        {
+            int years = totalMonths / 12;
+            int months = totalMonths % 12;
+            return $"{years}-{months}";
+        }
+
+        /// <summary>
+        /// Formats a duration value as "D HH:MM:SS.nnnnnnnnn", matching the Thrift output.
+        /// Example: 3 days + 12 h + 30 min + 15 s 500 ms → "3 12:30:15.500000000"
+        /// The <paramref name="unit"/> determines how to interpret <paramref name="rawValue"/>.
+        /// </summary>
+        internal static string FormatDuration(long rawValue, TimeUnit unit)
+        {
+            // Convert to nanoseconds first for uniform handling
+            long nanoseconds = unit switch
+            {
+                TimeUnit.Second => rawValue * 1_000_000_000L,
+                TimeUnit.Millisecond => rawValue * 1_000_000L,
+                TimeUnit.Microsecond => rawValue * 1_000L,
+                TimeUnit.Nanosecond => rawValue,
+                _ => rawValue * 1_000L // default to microseconds
+            };
+
+            // Separate the nanosecond sub-second part from whole seconds
+            long wholeSeconds = nanoseconds / 1_000_000_000L;
+            long subNanos = nanoseconds % 1_000_000_000L;
+
+            // Handle negative values: keep subNanos non-negative
+            if (subNanos < 0)
+            {
+                wholeSeconds -= 1;
+                subNanos += 1_000_000_000L;
+            }
+
+            long days = wholeSeconds / 86400L;
+            long remainderSeconds = wholeSeconds % 86400L;
+
+            // Handle negative days: ensure remainderSeconds is non-negative
+            if (remainderSeconds < 0)
+            {
+                days -= 1;
+                remainderSeconds += 86400L;
+            }
+
+            int hours = (int)(remainderSeconds / 3600L);
+            int minutes = (int)((remainderSeconds % 3600L) / 60L);
+            int seconds = (int)(remainderSeconds % 60L);
+
+            return $"{days} {hours:D2}:{minutes:D2}:{seconds:D2}.{subNanos:D9}";
+        }
 
         // --- JSON serialization helpers ---
 
--- a/csharp/src/ComplexTypeSerializingStream.cs
+++ b/csharp/src/ComplexTypeSerializingStream.cs
@@ -23,6 +23,7 @@ using Apache.Arrow;
 using Apache.Arrow.Adbc.Extensions;
 using Apache.Arrow.Ipc;
 using Apache.Arrow.Types;
+using AdbcDrivers.Databricks.StatementExecution;
 
 namespace AdbcDrivers.Databricks
 {
@@ -41,8 +42,6 @@ namespace AdbcDrivers.Databricks
     /// </summary>
     internal sealed class ComplexTypeSerializingStream : IArrowArrayStream
     {
-        private const string SparkSqlNameKey = "Spark:DataType:SqlName";
-
         private readonly IArrowArrayStream _inner;
         private readonly Schema _schema;
         private readonly HashSet<int> _complexColumnIndices;
@@ -105,7 +104,7 @@ namespace AdbcDrivers.Databricks
             {
                 Field field = schema.FieldsList[i];
                 if (field.Metadata != null &&
-                    field.Metadata.TryGetValue(SparkSqlNameKey, out string? sqlName) &&
+                    field.Metadata.TryGetValue(ColumnMetadataHelper.ArrowMetadataKey, out string? sqlName) &&
                     sqlName != null &&
                     (sqlName.Equals("ARRAY", StringComparison.OrdinalIgnoreCase) ||
                      sqlName.Equals("MAP", StringComparison.OrdinalIgnoreCase) ||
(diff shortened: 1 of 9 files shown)


csharp/src/StatementExecution/StatementExecutionStatement.cs
```diff
@@ -545,9 +552,14 @@ private IArrowType MapDatabricksTypeToArrowType(string typeName)
                 "DATE" => Date32Type.Default,
                 "TIMESTAMP" or "TIMESTAMP_NTZ" or "TIMESTAMP_LTZ" => TimestampType.Default,
                 "INTERVAL" => StringType.Default, // Intervals as strings for now
-                "ARRAY" => StringType.Default, // Complex types as strings for now
-                "MAP" => StringType.Default,
-                "STRUCT" => StringType.Default,
+                // Complex types use placeholder Arrow types so ComplexTypeSerializingStream
```
eric-wang-1990: This is not needed, this PR only focus on interval, for complex type we should tackle later

csharp/src/IntervalSerializingStream.cs
```diff
@@ -0,0 +1,228 @@
...
+            return builder.Build();
+        }
+
+        // --- Formatting helpers ---
+
+        /// <summary>
+        /// Formats a year-month interval (total months) as "Y-M", matching the Thrift output.
+        /// Example: 30 months → "2-6"
+        /// </summary>
+        internal static string FormatYearMonth(int totalMonths)
```
msrathore-db, line 175: **[F1 — Critical] Negative year-month interval formatted incorrectly**

For `totalMonths = -30`, C# integer division/modulo truncate toward zero: `years = -2`, `months = -6`, formatted as `"-2--6"`.

**Server returns** `"-2-6"` (verified via SEA REST API). **JDBC `IntervalConverter.formatYearMonth` returns** `"-2-6"` — uses `Math.abs(totalMonths)` then prepends a single `-`.

This is silent data corruption for any negative year-month interval and contradicts the PR's stated goal of matching Thrift behavior. The unit test at `csharp/test/Unit/IntervalSerializingStreamTests.cs:64` pins `"-2--6"` with the comment `// negative (mirrors Java driver)` — that comment is provably false; see the JDBC source at `IntervalConverter.java:85-94`.

**Fix** — mirror JDBC:
```csharp
bool neg = totalMonths < 0;
long abs = Math.Abs((long)totalMonths);  // (long) cast guards int.MinValue
long years = abs / 12;
long months = abs % 12;
return (neg ? "-" : "") + $"{years}-{months}";
```

csharp/test/Unit/IntervalSerializingStreamTests.cs
```diff
@@ -0,0 +1,326 @@
...
+        // FormatYearMonth static helper
+        // -----------------------------------------------------------------------
+
+        [Theory]
+        [InlineData(0, "0-0")]          // zero
+        [InlineData(12, "1-0")]         // exactly 1 year
+        [InlineData(30, "2-6")]         // 2 years 6 months
+        [InlineData(1, "0-1")]          // 1 month only
+        [InlineData(25, "2-1")]         // 2 years 1 month
+        [InlineData(-30, "-2--6")]      // negative (mirrors Java driver)
```
msrathore-db: **[F1 — Critical] Test pins wrong output for negative year-month**

The expected value `"-2--6"` does not match what JDBC produces. JDBC's `IntervalConverter.formatYearMonth` (`databricks-jdbc:src/main/java/com/databricks/jdbc/api/impl/converters/IntervalConverter.java:85-94`) takes `Math.abs(totalMonths)` then prepends `-` once, producing `"-2-6"`. The SEA server also returns `"-2-6"` for the same interval (verified via SEA REST API).

**Fix**:
```csharp
[InlineData(-30, "-2-6")]  // matches JDBC and server
[InlineData(-22, "-1-10")]
[InlineData(-10, "-0-10")]
[InlineData(-1, "-0-1")]
[InlineData(int.MinValue, ...)]  // overflow boundary
```
And drop the misleading "mirrors Java driver" comment.

csharp/src/IntervalSerializingStream.cs
```diff
@@ -0,0 +1,228 @@
...
+            int months = totalMonths % 12;
+            return $"{years}-{months}";
+        }
+
+        /// <summary>
+        /// Formats a duration value as "D HH:MM:SS.nnnnnnnnn", matching the Thrift output.
+        /// Example: 3 days + 12 h + 30 min + 15 s 500 ms → "3 12:30:15.500000000"
+        /// The <paramref name="unit"/> determines how to interpret <paramref name="rawValue"/>.
+        /// </summary>
+        internal static string FormatDuration(long rawValue, TimeUnit unit)
```
msrathore-db, line 189: **[F2 — Critical] Negative day-time interval formatted incorrectly**

For `rawValue = -304_215_000_000L` (microseconds; semantically `-3 days 12:30:15`), the floor-style "carry" arithmetic at lines 184–220 (`days -= 1; remainderSeconds += 86400`) emits `"-4 11:29:45.000000000"`.

**Server returns** `"-3 12:30:15"` (verified via SEA REST API). **JDBC** returns `"-3 12:30:15.000000000"` — negates upfront, formats positively, prepends `-`. See `IntervalConverter.java:99-124`.

The two outputs are mathematically equivalent decompositions of the same duration but the textual form differs. The server emits the JDBC form; this driver emits a different form, breaking Thrift parity. The existing Thrift-only `csharp/test/E2E/DateTimeValueTests.cs:82-105` already enumerates the correct server output — that contract is being violated by SEA.

**Fix** — mirror JDBC:
```csharp
long nanoseconds = unit switch { ... };
if (nanoseconds == long.MinValue) nanoseconds += 1;  // matches JDBC
bool neg = nanoseconds < 0;
if (neg) nanoseconds = -nanoseconds;
long days = nanoseconds / 86_400_000_000_000L;
// ... divmod on the absolute value
return (neg ? "-" : "") + $"{days} {hours:D2}:{minutes:D2}:{seconds:D2}.{subNanos:D9}";
```

csharp/test/E2E/IntervalValueTests.cs
```diff
@@ -0,0 +1,300 @@
+/*
```
msrathore-db, line 1: **[F3 — High] No negative-interval E2E coverage**

INTERVAL-001 through INTERVAL-010 cover only positive + zero + null. The cross-protocol parity tests INTERVAL-009 / INTERVAL-010 (which compare Thrift vs SEA cell-by-cell) only test positive values, so they cannot detect the divergence in F1/F2 — the most common bug class for any sign-handling code. Adding `INTERVAL '-2-6' YEAR TO MONTH` to INTERVAL-009 would have surfaced F1 immediately.

**Fix**:
- Extend INTERVAL-009 with `INTERVAL '-2-6'`, `INTERVAL '-0-10'`, `INTERVAL -178956970 YEAR`.
- Extend INTERVAL-010 with `INTERVAL '-3 12:30:15'`, `INTERVAL '-0 00:00:00.000001'`, `INTERVAL -106751991 DAYS 23 HOURS 59 MINUTES 59.999999 SECONDS`.
- Reuse the contract from `csharp/test/E2E/DateTimeValueTests.cs:82-105`, which already enumerates the correct server strings (currently Thrift-only via `Skip.If(Protocol == "rest", ...)` at line 109 — re-enable once F1/F2 are fixed).

csharp/src/ComplexTypeSerializingStream.cs
```diff
@@ -86,34 +109,29 @@ private static StringArray SerializeToStringArray(IArrowArray array)
...
-            for (int i = 0; i < original.FieldsList.Count; i++)
+            for (int i = 0; i < schema.FieldsList.Count; i++)
             {
-                Field field = original.FieldsList[i];
-                if (IsComplexType(field.DataType))
+                Field field = schema.FieldsList[i];
+                if (field.Metadata != null &&
+                    field.Metadata.TryGetValue(ColumnMetadataHelper.ArrowMetadataKey, out string? sqlName) &&
+                    sqlName != null &&
+                    (sqlName.Equals("ARRAY", StringComparison.OrdinalIgnoreCase) ||
```
msrathore-db: **[F5 — High] Detection silently fails for parameterised SqlName** (verified at runtime)

`sqlName.Equals("ARRAY", OrdinalIgnoreCase)` only matches the bare base name. I built a runtime test:

| Input metadata | Inner array | After `ComplexTypeSerializingStream` |
|---|---|---|
| `Spark:DataType:SqlName = "ARRAY<INT>"` | `ListArray` | `ListArray` (leaks through) |

The `ListArray` is not converted to a JSON `StringArray`; downstream consumers expecting strings break.

This works today on SEA only because [F11] collapses parameterised types to `"ARRAY"` via `SqlTypeNameParser`. **If F11 is fixed by passing raw `column.TypeText`, this exact-match detector silently breaks.** It also breaks on Thrift where the server may embed parameterised forms directly in the IPC field metadata.

**Fix**: either revert this file to Arrow-type detection (`type is ListType || MapType || StructType`) — aligns with the author's outstanding review comment that complex-type changes are out of scope — or use `StartsWith` matching with an Arrow-type fallback when metadata is absent.

csharp/src/StatementExecution/StatementExecutionStatement.cs
```diff
@@ -409,6 +416,11 @@ private IArrowArrayStream CreateReader(ExecuteStatementResponse response, Cancel
                 return new EmptyArrowArrayStream();
             }
 
+            // Derive the output schema from the manifest once, for all paths.
+            // JDBC uses the manifest schema exclusively for both inline and CloudFetch results;
+            // the Arrow IPC bytes are only used for data extraction, not schema definition.
+            Schema manifestSchema = GetSchemaFromManifest(response.Manifest);
```
msrathore-db: **[F10 — High] Empty-result regression: DDL on SEA now throws**

This hoisted call to `GetSchemaFromManifest` runs **before** the empty-result branch. For any response with `manifest.schema.columns_count == 0`, it throws `AdbcException("Result manifest does not contain schema information")`.

**Verified on the live warehouse** — same query against `origin/main` and `pr-440`:

| Statement | `origin/main` (pre-PR) | `pr-440` |
|---|---|---|
| `CREATE OR REPLACE TEMPORARY VIEW v_pr440_f10 AS SELECT 1 AS x` | **SUCCEEDED** — empty stream, schema-columns=0 | **THREW** `AdbcException: Result manifest does not contain schema information` |
| `DROP VIEW IF EXISTS v_pr440_f10` | **SUCCEEDED** | **THREW** same |
| `DROP VIEW IF EXISTS v_pr440_f10_does_not_exist` | **SUCCEEDED** | **THREW** same |

Server response for all three: `manifest.schema.columns_count = 0`, `total_chunk_count = 0`, no `attachment`, no `external_links`.

Pre-PR fell into the empty-result `else` branch:
```csharp
Schema schema = TryGetSchemaFromManifest(response.Manifest) ?? new Schema.Builder().Build();
return new EmptyArrowArrayStream(schema);
```
The hoist did not preserve that fallback. Production blast radius: every customer using SEA + DDL.

**Fix** — move the call into the inline/CloudFetch branches where it's actually needed, restore the empty-branch fallback:
```csharp
if (hasExternalLinks)
    return CreateCloudFetchReader(response, GetSchemaFromManifest(response.Manifest));
else if (response.Result?.Attachment is { Length: > 0 })
    return new InlineArrowStreamReader(..., manifestSchema: GetSchemaFromManifest(response.Manifest));
else
{
    Schema schema = TryGetSchemaFromManifest(response.Manifest) ?? new Schema.Builder().Build();
    return new EmptyArrowArrayStream(schema);
}
```

csharp/src/StatementExecution/StatementExecutionStatement.cs
```diff
@@ -503,18 +539,13 @@ private Schema GetSchemaFromManifest(ResultManifest manifest)
...
-                // SqlName. That key is not read by any known consumer today, so we omit
-                // it here for now. Add it if a consumer requires it (PECO-2950).
+                // Embed Spark:DataType:SqlName so downstream stream wrappers and consumers
+                // (e.g. the PowerBI connector's AdjustNativeTypes) can read the SQL type name.
+                // Note: the Thrift server also sets "Spark:DataType:JsonType" alongside SqlName;
+                // that key is not read by any known consumer today, so we omit it here (PECO-2950).
                 var metadata = new Dictionary<string, string>
                 {
-                    ["Spark:DataType:SqlName"] = ColumnMetadataHelper.GetSparkSqlName(typeName)
+                    [ColumnMetadataHelper.ArrowMetadataKey] = ColumnMetadataHelper.GetSparkSqlName(typeName)
```
msrathore-db: **[F11 — Medium] `Spark:DataType:SqlName` metadata is lossy on SEA — Thrift parity regression** (verified)

The PR feeds `column.TypeName` (bare `"INTERVAL"`/`"ARRAY"` from JSON `type_name`) into `GetSparkSqlName`. Even if the PR fed `column.TypeText` (full `"INTERVAL YEAR TO MONTH"`/`"ARRAY<INT>"`), `GetSparkSqlName` collapses parameterised types via `SqlTypeNameParser.BaseTypeName`. Verified at runtime:
```
GetSparkSqlName("INTERVAL YEAR TO MONTH") = "INTERVAL"
GetSparkSqlName("ARRAY<INT>")             = "ARRAY"
GetSparkSqlName("STRUCT<a:INT,b:STRING>") = "STRUCT"
GetSparkSqlName("MAP<STRING,INT>")        = "MAP"
```
**Verified on the live SEA server**, both fields are present in the manifest:
```
type_text="INTERVAL YEAR TO MONTH", type_name="INTERVAL"
type_text="ARRAY<INT>",             type_name="ARRAY"
```
The Thrift server's IPC-embedded `Spark:DataType:SqlName` carries the **full** `type_text`. The new SEA path embeds the **bare** `type_name`. Downstream consumers (PowerBI connector, etc.) read this metadata and lose information vs Thrift — ironic, since the design comment promotes this metadata as the cross-protocol detection signal.

**Fix**: special-case INTERVAL/ARRAY/MAP/STRUCT in `GetSparkSqlName` to return the original raw `column.TypeName`/`TypeText`, OR plumb the raw name into the metadata directly here at line 548.

csharp/src/StatementExecution/StatementExecutionStatement.cs
```diff
@@ -769,9 +810,17 @@ public InlineArrowStreamReader(
...
                 _memoryStream = new System.IO.MemoryStream(allData);
                 _streamReader = new ArrowStreamReader(_memoryStream);
+                // Use the manifest schema as the reported schema rather than the IPC-embedded schema.
+                // Arrow IPC files are self-describing, but the IPC schema carries native types
+                // (YearMonthIntervalType, ListType, etc.) while callers expect the output contract
+                // declared by the manifest (StringType for interval and complex columns). Keeping
+                // inline consistent with CloudFetch — which already uses GetSchemaFromManifest —
+                // lets IntervalSerializingStream and ComplexTypeSerializingStream detect and convert
+                // columns uniformly via Spark:DataType:SqlName metadata on every path.
+                _schema = manifestSchema;
```
msrathore-db, line 880: **[F6 — High] Manifest schema vs IPC schema mismatch is silent** (verified at runtime)

`InlineArrowStreamReader.Schema` reports `manifestSchema` externally, but `ReadNextRecordBatchAsync` returns IPC-decoded batches with their own schema. There is **no validation** that they agree. I verified at runtime — Apache.Arrow C# `RecordBatch` ctor accepts:

1. **Type mismatch**: schema declares `StringType` for column 0, array is `Int64Array` — no exception, no warning.
2. **Column-count mismatch**: schema declares 2 fields, only 1 array provided — no exception, no warning. This is even worse than the original review claim suggested.

In the happy path against the cooperating production server today, no divergence fires (verified live). But this is a latent silent-data-corruption surface: any future server bug, schema-evolution drift, or manifest/IPC version skew produces wrong output with zero diagnostic signal.

**Fix** (cheap, prevents a class of silent bugs): in `InlineArrowStreamReader` ctor, after `_streamReader = new ArrowStreamReader(_memoryStream)`, validate:
- `manifestSchema.FieldsList.Count == _streamReader.Schema.FieldsList.Count`
- For each non-converted column, field name and type match positionally.

Throw `AdbcException` on mismatch with both schemas serialized. Converts a class of silent-corruption bugs into loud failures.

csharp/src/StatementExecution/StatementExecutionStatement.cs
```diff
@@ -544,12 +583,12 @@ private IArrowType MapDatabricksTypeToArrowType(string typeName)
...
                 "DATE" => Date32Type.Default,
                 "TIMESTAMP" or "TIMESTAMP_NTZ" or "TIMESTAMP_LTZ" => TimestampType.Default,
-                "INTERVAL" => StringType.Default, // Intervals as strings for now
-                "ARRAY" => StringType.Default, // Complex types as strings for now
-                "MAP" => StringType.Default,
-                "STRUCT" => StringType.Default,
+                // Converted to string by IntervalSerializingStream; StringType is the output contract.
+                "INTERVAL" => StringType.Default,
+                // Converted to JSON string by ComplexTypeSerializingStream; StringType is the output contract.
+                "ARRAY" or "MAP" or "STRUCT" => StringType.Default,
```
msrathore-db, line 636: **[F12 — Medium] `Split('(')[0]` doesn't strip type qualifiers — explicit arms are dead code** (verified)

Verified at runtime:
```
"INTERVAL DAY TO SECOND".Split('(')[0].ToUpperInvariant() = "INTERVAL DAY TO SECOND"
"ARRAY<INT>".Split('(')[0].ToUpperInvariant()             = "ARRAY<INT>"
"MAP<STRING,INT>".Split('(')[0].ToUpperInvariant()        = "MAP<STRING,INT>"
"STRUCT<a:INT>".Split('(')[0].ToUpperInvariant()          = "STRUCT<A:INT>"
"DECIMAL(10,2)".Split('(')[0].ToUpperInvariant()          = "DECIMAL"
```
Only types that use parens (`DECIMAL(N,M)`, `VARCHAR(N)`) actually get stripped. Real server inputs like `"INTERVAL DAY TO SECOND"` and `"ARRAY<INT>"` (verified above; the server's `column.TypeText` field) fall through to the default arm `_ => StringType.Default`. The explicit `"INTERVAL"`, `"ARRAY"`, `"MAP"`, `"STRUCT"` arms are **dead code** — they happen to produce the same answer as the default. If anyone ever changes the default to `BinaryType` or `throw`, every interval/complex column silently breaks.

**Fix**: use `ColumnMetadataHelper.GetBaseTypeName(typeName)` (which goes through the proper `SqlTypeNameParser`) instead of `Split('(')[0]`:
```csharp
var baseType = ColumnMetadataHelper.GetBaseTypeName(typeName);
```

msrathore-db:
# Code Review Squad Report

**Merge Safety Score: 18/100 — CRITICAL RISK**

## Executive Summary

This PR ships **two silent data-correctness bugs for negative intervals on the SEA path** that contradict the PR's own stated goal of matching Thrift — verified against the live SEA server (`-2-6`, `-3 12:30:15`) and the JDBC reference (`Math.abs` then prepend `-`). The new C# unit tests **lock in** the wrong outputs (`-2--6`, `-4 11:29:45.000000000`) with a comment claiming they "mirror Java driver" — they don't. The new E2E suite has zero negative-interval cases, so the bug is invisible in CI; meanwhile an existing Thrift-only test (`csharp/test/E2E/DateTimeValueTests.cs:82-105`) already enumerates the correct server-produced strings and could be reused as the parity contract.

Additionally, the schema-everywhere refactor introduced a **proven production regression**: DDL statements (`CREATE/DROP VIEW`, etc.) on SEA now throw `AdbcException("Result manifest does not contain schema information")` — verified live on the test warehouse against both `origin/main` (succeeds) and this PR (throws).

## Inline findings (9 anchored comments above)

| # | Severity | Topic | File:line |
|---|---|---|---|
| F1 | Critical | Negative year-month formatted as `"-2--6"` (server: `"-2-6"`) | `IntervalSerializingStream.cs:176` + test pin at `IntervalSerializingStreamTests.cs:64` |
| F2 | Critical | Negative day-time formatted as `"-4 11:29:45"` (server: `"-3 12:30:15"`) | `IntervalSerializingStream.cs:188` |
| F3 | High | Zero negative-interval E2E cases; cross-protocol parity tests can't catch F1/F2 | `IntervalValueTests.cs` (file-level) |
| F5 | High | `ComplexTypeSerializingStream` exact-match detection silently fails on `"ARRAY<INT>"` (verified runtime) | `ComplexTypeSerializingStream.cs:125` |
| F6 | High | `RecordBatch` ctor accepts schema/array type AND column-count mismatches silently — no defensive guard (verified runtime) | `StatementExecutionStatement.cs:820` |
| F10 | High | DDL on SEA now throws — `CREATE/DROP VIEW` succeed on `origin/main`, throw on PR-440 (verified live warehouse) | `StatementExecutionStatement.cs:422` |
| F11 | Medium | `Spark:DataType:SqlName` metadata is lossy on SEA — Thrift parity regression (verified live + runtime) | `StatementExecutionStatement.cs:548` |
| F12 | Medium | `Split('(')[0]` doesn't strip qualifiers; explicit `INTERVAL/ARRAY/MAP/STRUCT` arms are dead code (verified runtime) | `StatementExecutionStatement.cs:589` |

## Additional medium findings (not file:line-anchorable)

- **F8** — `FormatDuration` silently overflows for `TimeUnit.Second` when `|rawValue| > 9.22e9` (right at Spark's documented day-time max). JDBC explicitly handles `Long.MIN_VALUE` and warns at `Long.MAX_VALUE`. After fixing F2, `Math.Abs(long.MinValue)` will throw — the same JDBC `+= 1` guard is needed. Wrap multiplications in `checked { }`.
- **F9** — No kill switch / feature flag for `IntervalSerializingStream`. Every SEA query is unconditionally wrapped; if a latent bug surfaces, customers can't work around it without a driver downgrade. The complex wrapper at least gates on `_enableComplexDatatypeSupport`. Add `databricks.serialize_intervals_to_string` defaulting to `true`.
- **F13** — Cancellation token forwarded to inner stream but never checked inside the per-row formatting loops in `SerializeIntervalToStringArray`. A 100k-row batch with several interval columns is uninterruptible mid-conversion. Plumb the token; check every ~4096 rows.
- **F14** — Zero observability: no log, no metric, no `Activity` event in either new file. By contrast, `StatementExecutionStatement.cs` already uses `Activity.Current?.AddEvent(...)` for `statement.set_option.unrecognized`, `statement.results_truncated`, etc. Ops investigating "interval value is wrong" gets no breadcrumbs.

## Verification methodology

Each finding was verified empirically:
- **Server ground truth**: queried the live SEA REST API directly. `INTERVAL '-2-6' YEAR TO MONTH` → server returns `"-2-6"`. `INTERVAL '-3 12:30:15' DAY TO SECOND` → server returns `"-3 12:30:15"`.
- **JDBC reference**: traced `IntervalConverter.formatYearMonth(-30)` and `formatFullDayTime(Duration.ofSeconds(-304215))` line-by-line. Both produce the server's format.
- **F4/F5/F6/F11/F12**: built a runtime test harness using the existing test project's `InternalsVisibleTo`, exercised the suspect code paths, asserted observed behavior. All claims confirmed.
- **F10**: ran the same DDL statements through the C# driver on `origin/main` and `pr-440`. `origin/main` succeeded for all three; PR-440 threw for all three. Clean regression.
- **F6 happy-path E2E**: ran `SELECT INTERVAL '2-6' YEAR TO MONTH, ARRAY(1,2), TIMESTAMP '2026-01-01'` against the live SEA server. No divergence today — F6 is a latent vulnerability, not an active bug.

## Bottom line

Block on **F1, F2, F10**. The negative-interval bugs corrupt data and contradict the PR's stated goal. F10 breaks DDL on SEA in production. F3 is a test-coverage gap that allowed F1/F2 to ship. F11/F12/F14 are bundled into the same PR but should be addressed.

Per the author's own outstanding review comment ("only focus on interval, complex type later"), consider splitting:
- **PR-A**: manifest-schema-for-all-paths refactor + complex detection rewrite
- **PR-B**: interval support (this PR's actual feature, with F1/F2 fixed)

---
*Review by [Code Review Squad](https://plugin-marketplace-1965859176160743.aws.databricksapps.com/plugin/code-review-squad). Feedback in [#code-review-squad-feedback](https://databricks.slack.com/channels/code-review-squad-feedback).*

csharp/src/StatementExecution/StatementExecutionStatement.cs
eric-wang-1990: ✅ Acknowledged — kept the scope to interval-only. The `ComplexTypeSerializingStream` wrapper remains (removing it would break schema/data consistency since the manifest schema already declares complex columns as `StringType`), but I made only the minimal fix needed: changed `Equals` → `StartsWith` in the detection logic to handle parameterised type names like `"ARRAY<INT>"`. No new complex-type functionality added.

---

csharp/src/IntervalSerializingStream.cs
```diff
@@ -0,0 +1,228 @@
...
+* limitations under the License.
+*/
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Apache.Arrow;
+using Apache.Arrow.Ipc;
+using Apache.Arrow.Scalars;
```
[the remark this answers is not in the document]
eric-wang-1990: ✅ Done — removed `using Apache.Arrow.Scalars;` from `IntervalSerializingStream.cs`.

---

csharp/src/IntervalSerializingStream.cs
eric-wang-1990: ✅ Fixed — `FormatYearMonth` now mirrors JDBC: negate upfront with `Math.Abs((long)totalMonths)`, format the absolute years/months, prepend `-` once. The `(long)` cast also guards `int.MinValue` overflow. `-30 months → "-2-6"` now matches server/JDBC.

---

csharp/test/Unit/IntervalSerializingStreamTests.cs
eric-wang-1990: ✅ Fixed — corrected the test expected values. `[InlineData(-30, "-2-6")]` now matches JDBC/server. Also added `[InlineData(-22, "-1-10")]`, `[InlineData(-10, "-0-10")]`, `[InlineData(-1, "-0-1")]` and dropped the misleading "mirrors Java driver" comment.

---

csharp/src/IntervalSerializingStream.cs
eric-wang-1990: ✅ Fixed — `FormatDuration` now mirrors JDBC: negate upfront, format the absolute value, prepend `-`. The floor-carry approach is gone. For `-304_215_000_000` µs: `neg=true`, `abs=304_215_000_000_000 ns` → `3 12:30:15.000000000` → `"-3 12:30:15.000000000"`. Also guards `long.MinValue` overflow before negation. The `FormatDuration_Microseconds_ReturnsExpectedString(-304215000000, "-3 12:30:15.000000000")` test now passes.

---

csharp/test/E2E/IntervalValueTests.cs
eric-wang-1990: ✅ Done — extended INTERVAL-009 with `INTERVAL '-2-6' YEAR TO MONTH` and `INTERVAL '-0-10' YEAR TO MONTH`, and INTERVAL-010 with `INTERVAL '-3 12:30:15' DAY TO SECOND` and `INTERVAL '-0 00:00:00.000001' DAY TO SECOND`. Both cross-protocol tests now cover negative values.

---

csharp/src/ComplexTypeSerializingStream.cs
eric-wang-1990: ✅ Fixed — changed `Equals("ARRAY")` / `Equals("MAP")` / `Equals("STRUCT")` to `StartsWith("ARRAY")` / `StartsWith("MAP")` / `StartsWith("STRUCT")` so parameterised forms like `"ARRAY<INT>"` are correctly detected. This is the minimal fix keeping complex-type handling out of the main feature scope.

---

csharp/src/StatementExecution/StatementExecutionStatement.cs
eric-wang-1990: ✅ Fixed — moved `GetSchemaFromManifest` out of the hoisted position and into the CloudFetch/inline branches where data actually exists. The empty-result (`else`) branch now uses `TryGetSchemaFromManifest(...) ?? new Schema.Builder().Build()`, so DDL statements with `columns_count=0` return an empty stream instead of throwing.

---

csharp/src/StatementExecution/StatementExecutionStatement.cs
eric-wang-1990: ✅ Fixed — `TryGetSchemaFromManifest` now uses `column.TypeText ?? GetSparkSqlName(column.TypeName)` for the `Spark:DataType:SqlName` metadata. `TypeText` carries the full form (e.g. `"INTERVAL YEAR TO MONTH"`, `"ARRAY<INT>"`) matching what the Thrift server embeds, with fallback to bare TypeName when TypeText is absent.

---

csharp/src/StatementExecution/StatementExecutionStatement.cs
eric-wang-1990: ✅ Done — added field-count validation in `InlineArrowStreamReader` ctor after loading the IPC stream. If `manifestSchema.FieldsList.Count != _streamReader.Schema.FieldsList.Count`, it throws `AdbcException` with both field lists serialised. Type mismatches (expected — manifest uses `StringType` for interval/complex columns) are not validated, but count divergence — the worst class of silent corruption — now fails loudly.

---

csharp/src/StatementExecution/StatementExecutionStatement.cs
eric-wang-1990: ✅ Fixed — replaced `typeName.Split('(')[0].ToUpperInvariant()` with `ColumnMetadataHelper.GetBaseTypeName(typeName).ToUpperInvariant()` in `MapDatabricksTypeToArrowType`. This goes through `SqlTypeNameParser` and correctly handles qualified forms like `"INTERVAL DAY TO SECOND"` → `"INTERVAL"` and `"ARRAY<INT>"` → `"ARRAY"`, making the explicit switch arms live rather than dead.

---

msrathore-db (approved):
Add some logging for easier debugging. LGTM

PR #441, opened 2026-05-05
[SEA] Fix GetCatalogsAsync to handle ADBC catalog patterns correctly (PECO-3018)

eric-wang-1990:
## Summary

Fixes the D8 bug case in `IGetObjectsDataProvider.GetCatalogsAsync` on the SEA (Statement Execution API) path.

**Root cause**: `GetCatalogsAsync` passed the ADBC `catalogPattern` directly to `ShowCatalogsCommand`, which converted it via `ConvertPattern()` and appended it as a SQL `LIKE` clause. This was wrong because:

- `"comparator\_tests"` → `SHOW CATALOGS LIKE 'comparator_tests'`  
  SQL `_` is a wildcard, so the escaped literal underscore was lost. The catalog `comparator_tests` would be found but so would `comparatorXtests` (too broad).
- `"%"` → `SHOW CATALOGS LIKE '*'` — glob `*` is not valid SQL LIKE syntax.
- `""` → `SHOW CATALOGS LIKE ''` — happened to return nothing, but by accident.

**Fix**: Mirror the Thrift path (`HiveServer2Connection.GetCatalogsAsync`) which fetches all catalogs via a bare `SHOW CATALOGS` request and filters results client-side using a pattern-to-regex conversion:

| `catalogPattern` | behavior |
|---|---|
| `null` | `SHOW CATALOGS`, no filter — return all (JDBC spec: null = any) |
| `""` | Return empty list immediately (no server call needed) |
| any other | `SHOW CATALOGS`, filter results with `CatalogPatternToRegex(pattern)` |

**New helpers** (analogous to `PatternToRegEx` in the Thrift layer):
- `ContainsUnescapedWildcard`: detects `%` or `_` not preceded by backslash
- `CatalogPatternToRegex`: converts ADBC/SQL pattern to anchored case-insensitive `Regex`
  - `%` → `.*`, `_` → `.`, `\_` → `_`, `\%` → `%`, `\\` → `\`

Also includes minor E2E test fixes to skip Thrift-specific tests when running against SEA-only endpoints.

**Note on hiveserver2**: This fix is entirely within the Databricks-specific layer (`csharp/src/StatementExecution/`). No changes were made to `csharp/hiveserver2/` (the upstream submodule).

## Test plan

- [x] 24 new unit tests in `GetCatalogsAsyncTests.cs` covering all pattern cases
- [x] `ContainsUnescapedWildcard`: 10 theory cases
- [x] `CatalogPatternToRegex`: wildcard expansion, escaped literals, case-insensitivity, regex-char escaping
- [x] D8 bug-report scenarios: `comparator\_tests`, `comparator_tests`, `%`, `compar%`, `""`, `nonexistent`
- [x] All unit tests pass: `dotnet test --filter "FullyQualifiedName~StatementExecution"` → 174 passed, 29 skipped (E2E only)
- [x] Build succeeds: `dotnet build AdbcDrivers.Databricks.csproj`

Fixes #PECO-3018

### DIFF
--- a/csharp/src/StatementExecution/StatementExecutionConnection.cs
+++ b/csharp/src/StatementExecution/StatementExecutionConnection.cs
@@ -18,6 +18,8 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net.Http;
+using System.Text;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using AdbcDrivers.Databricks.Http;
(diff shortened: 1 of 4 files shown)


eric-wang-1990:
Superseded by SEA catalog-pattern/wildcard fixes #536 (#525) + #594 (#593) and #604. Reopen if a gap remains.

PR #443, opened 2026-05-05
[SEA] Fix list_schemas to return empty rowset instead of throwing on non-matching catalog patterns (PECO-3017)

eric-wang-1990:
## Summary

- SEA's `GetObjects(depth=DbSchemas)` threw `DatabricksException` for any catalog argument that was not an exact-match against an existing catalog name, instead of returning an empty rowset.
- Affected inputs: `%`, `compar%`, `comparator\_tests`, `""`, `"nonexistent"`. Additionally `comparator_tests` (unescaped `_`) returned empty instead of matching via wildcard semantics.
- Root cause: `GetSchemasAsync()` passed the catalog pattern verbatim as a backtick-quoted SQL identifier, generating invalid SQL (e.g. `SHOW SCHEMAS IN \`%\``) that the server rejected.
- Fix: `GetSchemasAsync()` now classifies the catalog pattern and routes accordingly:
  1. `null` → `SHOW SCHEMAS IN ALL CATALOGS` (unchanged)
  2. Empty string → return empty rowset immediately
  3. Contains unescaped `%` or `_` → `SHOW SCHEMAS IN ALL CATALOGS` then filter rows client-side using a regex built from the ADBC pattern
  4. Literal name (with possible `\_`/`\%` escapes) → decode escapes, execute `SHOW SCHEMAS IN \`catalog\``, catch `AdbcException` and return empty rowset for nonexistent catalogs

## Test plan

- [x] 29 unit tests added in `GetSchemasAsyncTests.cs` covering all six bug-report scenarios and helper methods
- [x] `dotnet build` passes — 0 errors
- [x] All 746 unit tests pass

Fixes PECO-3017

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/csharp/src/StatementExecution/StatementExecutionConnection.cs
+++ b/csharp/src/StatementExecution/StatementExecutionConnection.cs
@@ -18,6 +18,8 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net.Http;
+using System.Text;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using AdbcDrivers.Databricks.Http;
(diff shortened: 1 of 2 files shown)


eric-wang-1990:
Superseded by #604 (align SEA metadata error behavior with Thrift — return empty instead of throwing). Reopen if a list_schemas gap remains.

PR #447, opened 2026-05-07
fix(csharp): map OAuth 401/403 to Unauthorized in SEA path (PECO-3007) [SEA]

eric-wang-1990:
## Summary

- **Root cause**: `OAuthClientCredentialsProvider.RefreshTokenInternalAsync` caught the HTTP error from `EnsureSuccessStatusCode()` and re-threw a `DatabricksException` without an `AdbcStatusCode`, defaulting to `UnknownError`
- **Fix**: Check the HTTP status code directly before throwing, mapping 401/403 → `AdbcStatusCode.Unauthorized` and other errors → `AdbcStatusCode.UnknownError`
- **Test**: Remove the `Skip.If` guard on `CanDetectInvalidAuthentication` for the REST protocol; add unit tests for the three status code cases

## Why SEA-specific in practice

The fix is in shared code (`OAuthClientCredentialsProvider`), but the Jira ticket is labelled `[SEA]` because the E2E test configurations differ:
- **Thrift** test config uses PAT: the 401 never passes through `OAuthClientCredentialsProvider` 
- **SEA/REST** test config uses OAuth M2M: the invalid client secret fails at the `/oidc/v1/token` step, hitting the bug

## Test plan

- [ ] New unit tests in `OAuthClientCredentialsProviderTests`: 401 → `Unauthorized`, 403 → `Unauthorized`, 500 → `UnknownError`
- [ ] All 812 unit tests pass locally
- [ ] `CanDetectInvalidAuthentication` skip removed for SEA — will be verified in E2E CI run

Closes PECO-3007

### DIFF
--- a/csharp/src/Auth/OAuthClientCredentialsProvider.cs
+++ b/csharp/src/Auth/OAuthClientCredentialsProvider.cs
@@ -118,37 +118,41 @@ namespace AdbcDrivers.Databricks.Auth
             try
             {
                 response = await _httpClient.SendAsync(request, cancellationToken);
+
+                if (!response.IsSuccessStatusCode)
+                {
+                    string errorBody = await response.Content.ReadAsStringAsync();
+                    var statusCode = response.StatusCode switch
+                    {
+                        System.Net.HttpStatusCode.Unauthorized or System.Net.HttpStatusCode.Forbidden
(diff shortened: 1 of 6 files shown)


msrathore-db (approved):
LGTM

PR #450, opened 2026-05-07
feat(csharp): implement Cancel() for SEA protocol (PECO-2939)

eric-wang-1990:
## Summary

Implements \`Cancel()\` for the SEA (REST) protocol path and aligns timeout behaviour with Thrift (PECO-2939, PECO-3013).

**Cancel support (PECO-2939)**
- Adds \`Cancel()\` override to \`StatementExecutionStatement\`
- Adds a lock-protected \`_executeCts\` field; \`ExecuteQueryAsync\` and \`ExecuteUpdateAsync\` now create a linked \`CancellationTokenSource\` so \`Cancel()\` can interrupt the in-flight poll loop immediately
- \`Cancel()\` cancels the local CTS and calls \`CancelStatementAsync\` on the server best-effort (\`POST /statements/{id}/cancel\`)
- Previously \`Cancel()\` fell through to \`AdbcStatement.Cancel()\` which always throws \`NotImplemented\`

**TimeoutException alignment (PECO-3013)**
- SEA now throws \`TimeoutException\` (not \`AdbcException\`) on poll timeout, matching Thrift behaviour
- \`PollWithTimeoutAsync\` deadline path changed from \`throw new AdbcException\` → \`throw new TimeoutException\`
- \`ExecuteQueryAsync\` and \`ExecuteUpdateAsync\` wrap \`OperationCanceledException\` (when the caller did not cancel) in \`TimeoutException\`

**\`enable_direct_results=false\` → \`wait_timeout=0\` mapping**
- \`StatementExecutionConnection\` now reads \`adbc.databricks.enable_direct_results\`; when set to \`"false"\` it overrides \`_waitTimeoutSeconds\` to 0, forcing the server to return immediately with PENDING/RUNNING (same semantics as disabling direct results on Thrift)

## Test plan

- [x] \`CancelRunningOperation\` passes in SEA mode — proxy asserts \`CountSeaCallsAsync("CancelStatement") == 1\`
- [x] \`SimultaneousStatementCancellation\` passes in SEA mode — proxy asserts \`InRange(1, 3)\` CancelStatement calls
- [x] \`StatementTimeoutTest\` (8 cases) passes in SEA mode — asserts \`TimeoutException\` is thrown
- [x] Unit tests in \`StatementExecutionQueryTimeoutTests.cs\` updated: \`ThrowsAsync<TimeoutException>\` for timeout cases, \`ThrowsAsync<OperationCanceledException>\` for explicit caller cancellation
- [x] No regression on Thrift path — \`Cancel()\` override is only in \`StatementExecutionStatement\` (SEA)

Fixes PECO-2939, PECO-3013.

### DIFF
--- a/csharp/src/StatementExecution/StatementExecutionConnection.cs
+++ b/csharp/src/StatementExecution/StatementExecutionConnection.cs
@@ -224,6 +224,11 @@ namespace AdbcDrivers.Databricks.StatementExecution
             properties.TryGetValue(DatabricksParameters.ResultCompression, out _resultCompression);
 
             _waitTimeoutSeconds = PropertyHelper.GetIntPropertyWithValidation(properties, DatabricksParameters.WaitTimeout, 10);
+            if (properties.TryGetValue(DatabricksParameters.EnableDirectResults, out var directResults) &&
+                directResults.Equals("false", StringComparison.OrdinalIgnoreCase))
+            {
+                _waitTimeoutSeconds = 0;
+            }
             _pollingIntervalMs = PropertyHelper.GetPositiveIntPropertyWithValidation(properties, DatabricksParameters.PollingInterval, 1000);
 
             // Memory pooling
--- a/csharp/src/StatementExecution/StatementExecutionStatement.cs
+++ b/csharp/src/StatementExecution/StatementExecutionStatement.cs
@@ -702,6 +702,25 @@ namespace AdbcDrivers.Databricks.StatementExecution
                 throw new InvalidOperationException("SQL query is required");
             }
 
+            var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+            lock (_cancelLock) { _executeCts = cts; }
+            try
+            {
+                return await ExecuteUpdateInternalAsync(cts.Token).ConfigureAwait(false);
+            }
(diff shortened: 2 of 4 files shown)


csharp/src/StatementExecution/StatementExecutionConnection.cs
```diff
@@ -224,6 +224,11 @@ private StatementExecutionConnection(
             properties.TryGetValue(DatabricksParameters.ResultCompression, out _resultCompression);
 
             _waitTimeoutSeconds = PropertyHelper.GetIntPropertyWithValidation(properties, DatabricksParameters.WaitTimeout, 10);
+            if (properties.TryGetValue(DatabricksParameters.EnableDirectResults, out var directResults) &&
```
jadewang-db, line 227: why do we need this?

jadewang-db (approved):

csharp/src/StatementExecution/StatementExecutionConnection.cs
eric-wang-1990: This is feature parity with Thrift, where it can disable EnableDirectResults by passing in this flag, SEA can achieve this by setting wait_timeout to 0.

PR #456, opened 2026-05-14
feat(csharp): add enable_fast_metadata_query flag for DESC TABLE EXTENDED

jadewang-db:
## Summary

Adds opt-in `adbc.databricks.enable_fast_metadata_query` flag (default `false`) that pairs the new SQL modifier `STATIC ONLY` (runtime PR #198486) on `DESC TABLE EXTENDED AS JSON` with the matching off-WLM routing signal per protocol.

Both signals are required together — `STATIC ONLY` alone still queues through WLM; off-WLM routing alone still does the full metadata scan.

| Protocol | SQL emitted | Off-WLM signal | Conditions |
|---|---|---|---|
| Thrift | `DESC TABLE EXTENDED <t> STATIC ONLY AS JSON` | `RunAsync=false` on descStmt | flag set AND warehouse path (`/sql/1.0/warehouses/{id}` or `/sql/1.0/endpoints/{id}`). General-cluster paths: flag is ignored. |
| SEA | `DESC TABLE EXTENDED <t> STATIC ONLY AS JSON` | `x-databricks-sea-can-run-fully-sync: true` (already sent by `ExecuteMetadataSqlAsync`) | flag set. SEA is always-warehouse. |

## Implementation

- `DatabricksConnection.IsWarehousePath` matches the path regex (cached) — only used to gate the Thrift opt-in.
- `DatabricksStatement.RunAsyncOverride` is a new internal nullable bool; when non-null it takes precedence over the connection-level `RunAsyncInThrift` inside `SetStatementProperties`. The override is set on the per-call descStmt in `GetColumnsExtendedAsync` rather than mutating connection state.
- `StatementExecutionConnection.EnableFastMetadataQuery` parses the same flag for SEA.

## Fallback safety

`STATIC ONLY` is rejected with parse error `42601` (`INVALID_STATIC_ONLY_USAGE`) on servers without PR #198486. The existing `catch (HiveServer2Exception ex) when (ex.SqlState == "42601" || ex.SqlState == "20000")` in `GetColumnsExtendedAsync` already handles that — falls back to `GetColumns + GetPrimaryKeys + GetCrossReference`.

## Test plan

- [x] Added unit tests for the new parameter constant.
- [x] Added unit tests for `IsWarehousePath` covering `/sql/1.0/warehouses/{id}`, `/sql/1.0/endpoints/{id}` (with/without trailing slash, with/without query strings), `/sql/protocolv1/...` clusters, empty path, and the `AdbcOptions.Uri` fallback.
- [x] Added unit tests for `UseFastMetadataQuery` covering all four combinations of `(flag, warehouse-path)`.
- [x] Full `dotnet test --filter "FullyQualifiedName~Unit."` suite: 638 passed, 0 failed.
- [ ] E2E verification against a DBSQL warehouse running PR #198486 is gated on that runtime change reaching the test env; existing 42601 fallback protects pre-rollout.

Closes PECO-3021

### DIFF
--- a/csharp/src/DatabricksConnection.cs
+++ b/csharp/src/DatabricksConnection.cs
@@ -379,8 +379,6 @@ namespace AdbcDrivers.Databricks
         private static readonly System.Text.RegularExpressions.Regex s_warehousePathPattern =
             new System.Text.RegularExpressions.Regex(@"^/sql/1\.0/(warehouses|endpoints)/[^/]+/?$");
 
-        private bool? _isWarehousePathCached;
-
         /// <summary>
         /// True when the configured connection path targets a DBSQL warehouse
         /// (/sql/1.0/warehouses/{id} or /sql/1.0/endpoints/{id}). False for general
@@ -390,15 +388,13 @@ namespace AdbcDrivers.Databricks
         {
             get
             {
-                if (_isWarehousePathCached.HasValue)
-                {
-                    return _isWarehousePathCached.Value;
-                }
-
                 string? path = null;
                 if (Properties.TryGetValue(SparkParameters.Path, out string? rawPath) && !string.IsNullOrEmpty(rawPath))
                 {
                     path = rawPath;
+                    // Only the raw-Path branch can carry a query string; Uri.AbsolutePath strips it.
+                    int q = path!.IndexOf('?');
+                    if (q >= 0) path = path.Substring(0, q);
                 }
                 else if (Properties.TryGetValue(AdbcOptions.Uri, out string? uri)
                     && !string.IsNullOrEmpty(uri)
@@ -407,14 +403,7 @@ namespace AdbcDrivers.Databricks
                     path = parsedUri.AbsolutePath;
                 }
 
-                if (!string.IsNullOrEmpty(path))
-                {
-                    int q = path!.IndexOf('?');
-                    if (q >= 0) path = path.Substring(0, q);
-                }
-
-                _isWarehousePathCached = !string.IsNullOrEmpty(path) && s_warehousePathPattern.IsMatch(path!);
-                return _isWarehousePathCached.Value;
+                return !string.IsNullOrEmpty(path) && s_warehousePathPattern.IsMatch(path!);
             }
         }
 
--- a/csharp/src/DatabricksConnection.cs
+++ b/csharp/src/DatabricksConnection.cs
@@ -90,6 +90,7 @@ namespace AdbcDrivers.Databricks
         private const bool DefaultRateLimitRetry = true;
         private const bool DefaultTransportErrorRetry = true;
         private bool _useDescTableExtended = false;
+        private bool _enableFastMetadataQuery = false;
 
         // Trace propagation configuration
         private bool _tracePropagationEnabled = true;
@@ -207,6 +208,7 @@ namespace AdbcDrivers.Databricks
             _useCloudFetch = PropertyHelper.GetBooleanPropertyWithValidation(Properties, DatabricksParameters.UseCloudFetch, _useCloudFetch);
             _canDecompressLz4 = PropertyHelper.GetBooleanPropertyWithValidation(Properties, DatabricksParameters.CanDecompressLz4, _canDecompressLz4);
             _useDescTableExtended = PropertyHelper.GetBooleanPropertyWithValidation(Properties, DatabricksParameters.UseDescTableExtended, _useDescTableExtended);
+            _enableFastMetadataQuery = PropertyHelper.GetBooleanPropertyWithValidation(Properties, DatabricksParameters.EnableFastMetadataQuery, _enableFastMetadataQuery);
             _runAsyncInThrift = PropertyHelper.GetBooleanPropertyWithValidation(Properties, DatabricksParameters.EnableRunAsyncInThriftOp, _runAsyncInThrift);
             _enableComplexDatatypeSupport = PropertyHelper.GetBooleanPropertyWithValidation(Properties, DatabricksParameters.EnableComplexDatatypeSupport, _enableComplexDatatypeSupport);
 
@@ -373,6 +375,55 @@ namespace AdbcDrivers.Databricks
         /// Check if current connection can use `DESC TABLE EXTENDED` query
         /// </summary>
         internal bool CanUseDescTableExtended => _useDescTableExtended && ServerProtocolVersion != null && FeatureVersionNegotiator.SupportsDESCTableExtended(ServerProtocolVersion.Value);
+
+        private static readonly System.Text.RegularExpressions.Regex s_warehousePathPattern =
+            new System.Text.RegularExpressions.Regex(@"^/sql/1\.0/(warehouses|endpoints)/[^/]+/?$");
+
+        private bool? _isWarehousePathCached;
+
+        /// <summary>
+        /// True when the configured connection path targets a DBSQL warehouse
+        /// (/sql/1.0/warehouses/{id} or /sql/1.0/endpoints/{id}). False for general
+        /// clusters (/sql/protocolv1/o/{orgId}/{clusterId}) or when no path is set.
+        /// </summary>
+        internal bool IsWarehousePath
+        {
+            get
+            {
+                if (_isWarehousePathCached.HasValue)
+                {
+                    return _isWarehousePathCached.Value;
+                }
+
+                string? path = null;
+                if (Properties.TryGetValue(SparkParameters.Path, out string? rawPath) && !string.IsNullOrEmpty(rawPath))
+                {
+                    path = rawPath;
+                }
+                else if (Properties.TryGetValue(AdbcOptions.Uri, out string? uri)
+                    && !string.IsNullOrEmpty(uri)
+                    && Uri.TryCreate(uri, UriKind.Absolute, out Uri? parsedUri))
+                {
+                    path = parsedUri.AbsolutePath;
+                }
+
+                if (!string.IsNullOrEmpty(path))
+                {
+                    int q = path!.IndexOf('?');
+                    if (q >= 0) path = path.Substring(0, q);
+                }
+
+                _isWarehousePathCached = !string.IsNullOrEmpty(path) && s_warehousePathPattern.IsMatch(path!);
+                return _isWarehousePathCached.Value;
+            }
+        }
+
+        /// <summary>
+        /// True when the driver should opt into the fast metadata query path. Requires
+        /// both the connection flag and a DBSQL warehouse path; otherwise false.
+        /// See <see cref="DatabricksParameters.EnableFastMetadataQuery"/>.
+        /// </summary>
+        internal bool UseFastMetadataQuery => _enableFastMetadataQuery && IsWarehousePath;
 
         /// <summary>
         /// Gets whether PK/FK metadata call is enabled
--- a/csharp/src/DatabricksStatement.cs
+++ b/csharp/src/DatabricksStatement.cs
@@ -71,6 +71,14 @@ namespace AdbcDrivers.Databricks
         internal bool IsInternalCall { get; set; } // Marks if this is a driver-internal operation (e.g., USE SCHEMA)
 
         /// <summary>
+        /// Optional override for the Thrift RunAsync flag on a single statement. When non-null,
+        /// takes precedence over the connection-level <see cref="DatabricksConnection.RunAsyncInThrift"/>.
+        /// Pairs with the SQL-level STATIC ONLY modifier on DESC TABLE EXTENDED: RunAsync=false
+        /// is what tells the warehouse to route the command off the WLM path.
+        /// </summary>
+        internal bool? RunAsyncOverride { get; set; }
+
+        /// <summary>
         /// Telemetry context for the current statement execution, pending emission on Dispose.
         /// Set before calling base.ExecuteQueryAsync()/ExecuteQuery() so that
         /// <see cref="DatabricksConnection.NewReader{T}"/> can forward it to the
@@ -445,7 +453,7 @@ namespace AdbcDrivers.Databricks
             statement.CanDownloadResult = useCloudFetch;
             statement.CanDecompressLZ4Result = canDecompressLz4;
             statement.MaxBytesPerFile = maxBytesPerFile;
-            statement.RunAsync = runAsyncInThrift;
+            statement.RunAsync = RunAsyncOverride ?? runAsyncInThrift;
 
             Connection.TrySetGetDirectResults(statement);
 
@@ -471,7 +479,11 @@ namespace AdbcDrivers.Databricks
             Activity.Current?.SetTag("statement.cloudfetch.can_decompress_lz4", canDecompressLz4);
             Activity.Current?.SetTag("statement.cloudfetch.max_bytes_per_file", maxBytesPerFile);
             Activity.Current?.SetTag("statement.cloudfetch.max_bytes_per_file_mb", maxBytesPerFile / 1024.0 / 1024.0);
-            Activity.Current?.SetTag("statement.property.run_async", runAsyncInThrift);
+            Activity.Current?.SetTag("statement.property.run_async", statement.RunAsync);
+            if (RunAsyncOverride.HasValue)
+            {
+                Activity.Current?.SetTag("statement.property.run_async.source", "override");
+            }
 
             Activity.Current?.AddEvent("statement.set_properties.complete");
         }
@@ -993,13 +1005,26 @@ namespace AdbcDrivers.Databricks
                     return baseResult;
                 }
 
-                string query = $"DESC TABLE EXTENDED {fullTableName} AS JSON";
+                // Fast metadata: STATIC ONLY (runtime PR #198486) skips Delta log / Mesa RPCs,
+                // and RunAsync=false routes the request off the WLM path. Both signals are required —
+                // STATIC ONLY without RunAsync=false still goes through WLM; RunAsync=false without
+                // STATIC ONLY still does the full metadata scan.
+                bool useFastMetadataQuery = ((DatabricksConnection)Connection).UseFastMetadataQuery;
+                string query = useFastMetadataQuery
+                    ? $"DESC TABLE EXTENDED {fullTableName} STATIC ONLY AS JSON"
+                    : $"DESC TABLE EXTENDED {fullTableName} AS JSON";
                 activity?.AddEvent("statement.desc_table_extended.executing_query", [
-                    new("query_summary", query.Length > 100 ? query.Substring(0, 100) + "..." : query)
+                    new("query_summary", query.Length > 100 ? query.Substring(0, 100) + "..." : query),
+                    new("fast_metadata_query", useFastMetadataQuery)
                 ]);
 
                 using var descStmt = Connection.CreateStatement();
                 descStmt.SqlQuery = query;
+
+                if (useFastMetadataQuery && descStmt is DatabricksStatement databricksDescStmt)
+                {
+                    databricksDescStmt.RunAsyncOverride = false;
+                }
                 QueryResult descResult;
 
                 try
--- a/csharp/src/DatabricksStatement.cs
+++ b/csharp/src/DatabricksStatement.cs
@@ -984,7 +984,8 @@ namespace AdbcDrivers.Databricks
             {
                 activity?.AddEvent("statement.get_columns_extended.start");
                 string? fullTableName = BuildTableName();
-                var canUseDescTableExtended = ((DatabricksConnection)Connection).CanUseDescTableExtended;
+                var connection = (DatabricksConnection)Connection;
+                var canUseDescTableExtended = connection.CanUseDescTableExtended;
 
                 activity?.SetTag("statement.catalog_name", CatalogName ?? "(none)");
                 activity?.SetTag("statement.schema_name", SchemaName ?? "(none)");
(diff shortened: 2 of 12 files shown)


jadewang-db (commented):
## Self-review

Overall the change is small and well-scoped. The Thrift-side opt-in plus warehouse-path gating matches the JIRA, and SEA was correctly identified as already wired via `x-databricks-sea-can-run-fully-sync`. One high-priority PR-hygiene issue around the branch state, plus a handful of low-priority simplifications inline.

### Positive observations
- `UseFastMetadataQuery = _enableFastMetadataQuery && IsWarehousePath` is the right composition; unit tests enumerate the full truth table.
- Per-statement `RunAsyncOverride` is the right surface — mutating connection state would have been worse.
- Existing 42601/20000 fallback in `GetColumnsExtendedAsync` still protects against servers that reject the new shape, so rollout is safe.
- Activity tag `fast_metadata_query` will make the flag's effect easy to confirm in production traces.

### HIGH — branch is significantly behind `main`
The PR diff lists 11 files (+277/-9), but the commit `f8f66af` only touches 6 files (+200/-3). The extras come from telemetry refactors that landed on main while this branch was idle (e.g., PECO-2978/2988/2992 in `DatabricksStatement.cs` and `ConnectionTelemetry.cs`), plus `d02dae2` which was already merged via #401.

```
$ git log origin/main..PECO-3021 --oneline
f8f66af feat(csharp): add enable_fast_metadata_query flag for DESC TABLE EXTENDED
d02dae2 fix(csharp): emit bare hostname for telemetry host_url to match JDBC   ← already merged via #401
```

Recommend `git rebase origin/main` and force-push so the diff cleanly reflects only PECO-3021's work.

### Test coverage note
Coverage of new code is exclusively unit-level. Wire behavior (server actually receiving `RunAsync=false`) is not asserted — relies on the existing E2E harness once runtime PR #198486 deploys. Worth mentioning the intentional E2E gap in the PR body.

csharp/src/DatabricksStatement.cs
```diff
@@ -951,12 +962,22 @@ protected override async Task<QueryResult> GetColumnsExtendedAsync(CancellationT
...
+                    new("fast_metadata_query", useFastMetadataQuery)
                 ]);
 
                 using var descStmt = Connection.CreateStatement();
                 descStmt.SqlQuery = query;
+
+                // For DBSQL warehouses with the fast-metadata flag enabled, force RunAsync=false
+                // on this DESC TABLE EXTENDED statement so the server treats it as a metadata
+                // query and bypasses WLM. No effect on general clusters (flag gated by warehouse path).
+                if (useFastMetadataQuery && descStmt is DatabricksStatement databricksDescStmt)
```
jadewang-db: **MEDIUM** — The `descStmt is DatabricksStatement databricksDescStmt` defensive check can't actually fail today: `DatabricksConnection.CreateStatement()` always returns a `DatabricksStatement`. If a future refactor ever changes that return type, the override is silently dropped without telemetry.

Suggest either dropping the `is` and casting directly (the same cast already exists at L932), or adding an activity event when the type is unexpected. Simpler form:
```csharp
var databricksDescStmt = (DatabricksStatement)descStmt;
if (useFastMetadataQuery)
{
    databricksDescStmt.RunAsyncOverride = false;
}
```

csharp/src/DatabricksConnection.cs
```diff
@@ -363,6 +365,55 @@ protected internal override bool TrySetGetDirectResults(IRequest request)
         /// </summary>
         internal bool CanUseDescTableExtended => _useDescTableExtended && ServerProtocolVersion != null && FeatureVersionNegotiator.SupportsDESCTableExtended(ServerProtocolVersion.Value);
 
+        private static readonly System.Text.RegularExpressions.Regex s_warehousePathPattern =
+            new System.Text.RegularExpressions.Regex(@"^/sql/1\.0/(warehouses|endpoints)/[^/]+/?$");
+
+        private bool? _isWarehousePathCached;
```
jadewang-db: **MEDIUM** — `private bool? _isWarehousePathCached;` — writes to `Nullable<bool>` aren't atomic in .NET (two-field struct). The function is deterministic so the worst case is just recomputing once, but the cache adds complexity for a regex check that runs at most once per `GetColumnsExtended` call.

Suggest dropping the cache entirely — the static regex is the only caching that actually matters here:
```csharp
internal bool IsWarehousePath
{
    get
    {
        string? path = ResolveConnectionPath();
        return !string.IsNullOrEmpty(path) && s_warehousePathPattern.IsMatch(path!);
    }
}
```

csharp/src/DatabricksStatement.cs
```diff
@@ -951,12 +962,22 @@ protected override async Task<QueryResult> GetColumnsExtendedAsync(CancellationT
                 }
 
                 string query = $"DESC TABLE EXTENDED {fullTableName} AS JSON";
+                bool useFastMetadataQuery = ((DatabricksConnection)Connection).UseFastMetadataQuery;
```
jadewang-db: **LOW** — Duplicate `(DatabricksConnection)Connection` cast: this line and L932 both perform it. Extract a single local at the top of `GetColumnsExtendedAsync`:
```csharp
var connection = (DatabricksConnection)Connection;
var canUseDescTableExtended = connection.CanUseDescTableExtended;
// ...
bool useFastMetadataQuery = connection.UseFastMetadataQuery;
```

csharp/src/DatabricksStatement.cs
```diff
@@ -428,7 +435,11 @@ protected override void SetStatementProperties(TExecuteStatementReq statement)
             Activity.Current?.SetTag("statement.cloudfetch.can_decompress_lz4", canDecompressLz4);
             Activity.Current?.SetTag("statement.cloudfetch.max_bytes_per_file", maxBytesPerFile);
             Activity.Current?.SetTag("statement.cloudfetch.max_bytes_per_file_mb", maxBytesPerFile / 1024.0 / 1024.0);
-            Activity.Current?.SetTag("statement.property.run_async", runAsyncInThrift);
+            Activity.Current?.SetTag("statement.property.run_async", statement.RunAsync);
```
jadewang-db, line 482: **LOW** — Semantic change worth flagging: this tag previously reported the connection-level setting; now it reports the effective post-override value. Strict improvement for accuracy, but downstream dashboards keying on the old semantics will silently shift.

The new `statement.property.run_async.overridden` flag below is good. Consider replacing it with `statement.property.run_async.source = "override"|"connection"` for self-describing telemetry. Otherwise, just call out the semantic shift in the PR description.

csharp/src/DatabricksConnection.cs
```diff
@@ -363,6 +365,55 @@ protected internal override bool TrySetGetDirectResults(IRequest request)
         /// </summary>
         internal bool CanUseDescTableExtended => _useDescTableExtended && ServerProtocolVersion != null && FeatureVersionNegotiator.SupportsDESCTableExtended(ServerProtocolVersion.Value);
 
+        private static readonly System.Text.RegularExpressions.Regex s_warehousePathPattern =
```
jadewang-db, line 379: **LOW** — Regex pinned to `/sql/1.0/`. Same as `StatementExecutionConnection.cs:162`, so this is intentional consistency — but if a future `/sql/1.1/...` path ships, both spots will need to be updated together. A shared helper in `PropertyHelper` would prevent drift.

csharp/src/DatabricksStatement.cs
```diff
@@ -951,12 +962,22 @@ protected override async Task<QueryResult> GetColumnsExtendedAsync(CancellationT
...
-                    new("query_summary", query.Length > 100 ? query.Substring(0, 100) + "..." : query)
+                    new("query_summary", query.Length > 100 ? query.Substring(0, 100) + "..." : query),
+                    new("fast_metadata_query", useFastMetadataQuery)
                 ]);
 
                 using var descStmt = Connection.CreateStatement();
                 descStmt.SqlQuery = query;
+
+                // For DBSQL warehouses with the fast-metadata flag enabled, force RunAsync=false
+                // on this DESC TABLE EXTENDED statement so the server treats it as a metadata
```
jadewang-db: **NIT** — The first sentence (the *why* — server-side WLM bypass) is the useful part of this comment. The second sentence ("No effect on general clusters") restates what `useFastMetadataQuery` already encodes via the warehouse-path gate, and could be dropped.

csharp/src/DatabricksConnection.cs
```diff
@@ -363,6 +365,55 @@ protected internal override bool TrySetGetDirectResults(IRequest request)
...
+                    path = parsedUri.AbsolutePath;
+                }
+
+                if (!string.IsNullOrEmpty(path))
+                {
+                    int q = path!.IndexOf('?');
+                    if (q >= 0) path = path.Substring(0, q);
+                }
+
+                _isWarehousePathCached = !string.IsNullOrEmpty(path) && s_warehousePathPattern.IsMatch(path!);
```
jadewang-db: **Medium / concurrency** — `_isWarehousePathCached` is a `Nullable<bool>` written and read without synchronization. `Nullable<T>` is a two-field struct (`hasValue`, `value`), and reads/writes are not atomic. A racing reader could observe `HasValue == true` with the stale default `Value == false` before the write of `value` completes. Result is idempotent so the practical impact is minimal (worst case the cache reports `false` once), but worth fixing for cleanliness on a property that gates a code path.

Options:
- Drop the cache — the regex match is microseconds and `IsWarehousePath` is read at most once per `GetColumnsExtended` call.
- Use `Lazy<bool>` with `LazyThreadSafetyMode.ExecutionAndPublication`.
- Compute eagerly in the constructor / `OpenAsync` next to the other property reads (cleanest if you want the cache).

ADBC connections aren't formally documented as thread-safe in this driver, but other lazy fields here use eager init, so eager init would also match the local convention.

csharp/src/DatabricksConnection.cs
```diff
@@ -363,6 +365,55 @@ protected internal override bool TrySetGetDirectResults(IRequest request)
...
+                }
+                else if (Properties.TryGetValue(AdbcOptions.Uri, out string? uri)
+                    && !string.IsNullOrEmpty(uri)
+                    && Uri.TryCreate(uri, UriKind.Absolute, out Uri? parsedUri))
+                {
+                    path = parsedUri.AbsolutePath;
+                }
+
+                if (!string.IsNullOrEmpty(path))
+                {
```
jadewang-db: Minor: when the path comes from `AdbcOptions.Uri`, `parsedUri.AbsolutePath` already excludes the query string, so the `IndexOf('?')` strip a few lines below is dead code for that branch. It only matters for the `SparkParameters.Path` raw-string branch. Not wrong — just slightly misleading. A short comment ("only the raw-Path branch can carry a `?`; AbsolutePath already strips it") would make the intent clearer, or hoist the strip into the raw-Path branch only.

csharp/src/DatabricksStatement.cs
```diff
@@ -950,13 +961,26 @@ protected override async Task<QueryResult> GetColumnsExtendedAsync(CancellationT
...
                 activity?.AddEvent("statement.desc_table_extended.executing_query", [
-                    new("query_summary", query.Length > 100 ? query.Substring(0, 100) + "..." : query)
+                    new("query_summary", query.Length > 100 ? query.Substring(0, 100) + "..." : query),
+                    new("fast_metadata_query", useFastMetadataQuery)
                 ]);
 
                 using var descStmt = Connection.CreateStatement();
                 descStmt.SqlQuery = query;
+
+                if (useFastMetadataQuery && descStmt is DatabricksStatement databricksDescStmt)
```
jadewang-db: **Low / defensive** — the `descStmt is DatabricksStatement databricksDescStmt` pattern silently no-ops if the cast fails: the `STATIC ONLY` SQL still goes out, but without `RunAsync=false`, so we'd emit a fast-metadata query that still goes through WLM — the very scenario the PR description warns about ("STATIC ONLY alone still queues through WLM").

In the current code `DatabricksConnection.CreateStatement` always returns a `DatabricksStatement`, so this can't happen today. Two cheap hardening options:
- `Debug.Assert(descStmt is DatabricksStatement, ...)` so we'd catch a future refactor in test, or
- cast unconditionally (`var databricksDescStmt = (DatabricksStatement)descStmt;`) — if the invariant ever breaks, an `InvalidCastException` is a louder failure than a silent perf regression.

Not a blocker; flagging because the failure mode is invisible at runtime.

csharp/test/E2E/Telemetry/ConnectionParametersTests.cs
```diff
@@ -154,6 +154,72 @@ public async Task ConnectionParams_HostInfoPort_IsPopulated()
...
+                var protoLog = TelemetryTestHelpers.GetProtoLog(logs[0]);
+
+                Assert.NotNull(protoLog.DriverConnectionParams);
+                Assert.NotNull(protoLog.DriverConnectionParams.HostInfo);
+                string actualHostUrl = protoLog.DriverConnectionParams.HostInfo.HostUrl;
+
+                Assert.False(string.IsNullOrEmpty(actualHostUrl), "host_url should be populated");
+                Assert.Equal(expectedHost, actualHostUrl);
+                Assert.DoesNotContain("://", actualHostUrl);
+                Assert.DoesNotContain(":", actualHostUrl);
```
jadewang-db: `Assert.DoesNotContain(":", actualHostUrl)` is stricter than the spec. The new contract is "bare hostname, no scheme, no port suffix" — IPv6 literals (legal in `Uri.Host`) contain colons and would fail this assertion. Databricks workspace hostnames are always FQDNs in practice so this won't fire today, but if a customer ever uses a literal IP for testing, the regression test trips for the wrong reason.

Suggest dropping the bare `:` check and relying on `DoesNotContain("://", ...)` + `Equal(expectedHost, actualHostUrl)` above, which together already enforce no-scheme/no-port. Or scope it to `!actualHostUrl.StartsWith("[")` if you want to keep an explicit no-port check.

jadewang-db (commented):
Self-review pass on PECO-3021. Overall the design is clean — the AND-gate of flag + warehouse-path on Thrift, the per-statement `RunAsyncOverride` to avoid mutating connection state, and the reuse of the existing `42601`/`20000` fallback for old servers are all the right shapes. The unit-test matrix on `IsWarehousePath` and `UseFastMetadataQuery` is thorough.

Findings below are mostly Low/Medium polish — no Critical or High issues. One scope/PR-hygiene note: this PR also carries commit d02dae2 (telemetry host_url bare-hostname fix, PECO-2987) which the description doesn't mention. Consider either splitting it out or extending the PR body so reviewers know both changes are in flight.

csharp/src/DatabricksConnection.cs
jadewang-db: Done in latest commit — dropped `_isWarehousePathCached` entirely. `IsWarehousePath` now computes fresh on each call (regex match is microseconds, called at most once per `GetColumnsExtended`). Removes the `Nullable<bool>` torn-read concern and simplifies the getter.

csharp/src/DatabricksConnection.cs
jadewang-db: Done in same commit as the line-371 thread — cache dropped, getter recomputes on each call. The redundant `IndexOf(?)` strip on the URI branch was also scoped to the raw-Path branch (see thread on L400).

csharp/src/DatabricksConnection.cs
jadewang-db: Done — moved the `IndexOf(?)` strip inside the raw-Path branch where it is actually needed. The Uri branch now relies on `parsedUri.AbsolutePath` alone (it already strips the query string).

csharp/src/DatabricksStatement.cs
jadewang-db: Done — went with the unconditional cast option you mentioned. Same change addresses the older thread at L980; reviewers please see that one for details.

csharp/src/DatabricksStatement.cs
jadewang-db: Done — extracted `var connection = (DatabricksConnection)Connection;` at the top of `GetColumnsExtendedAsync`. Both `CanUseDescTableExtended` and `UseFastMetadataQuery` now go through it, and `CreateStatement` is called on the same local.

csharp/src/DatabricksStatement.cs
jadewang-db: Done — `descStmt` is now created via an unconditional `(DatabricksStatement)connection.CreateStatement()` cast. If a future refactor ever returns a non-`DatabricksStatement`, this throws `InvalidCastException` immediately instead of silently emitting a fast-metadata query that still goes through WLM.

csharp/src/DatabricksStatement.cs
jadewang-db: Outdated — that comment block was rewritten in the latest commit. The new comment at L964-965 is two sentences focused on the *why* (WLM bypass) and the "both signals required" invariant; the "no effect on general clusters" line is gone.

csharp/src/DatabricksConnection.cs
jadewang-db: Acknowledged — keeping the duplicated regex in this PR to scope the change tightly. A shared `PropertyHelper.IsWarehousePath(path)` (used by both `DatabricksConnection` and `StatementExecutionConnection`) makes sense; filing as follow-up rather than expanding this PR.

csharp/src/DatabricksStatement.cs
jadewang-db: Already implemented in the same commit — `statement.property.run_async` reports the effective (post-override) value, and a sibling `statement.property.run_async.source = "override"` tag is set when `RunAsyncOverride.HasValue` (DatabricksStatement.cs L439-442). I left the source tag absent (rather than `="connection"`) for the default case to keep the cardinality down; happy to set it explicitly if you prefer self-describing on both branches.

csharp/test/E2E/Telemetry/ConnectionParametersTests.cs
jadewang-db: Done — dropped `Assert.DoesNotContain(":", actualHostUrl)`. The remaining `Assert.Equal(expectedHost, actualHostUrl)` and `Assert.DoesNotContain("://", actualHostUrl)` together enforce the bare-hostname contract without rejecting IPv6 literals.

csharp/src/DatabricksStatement.cs
```diff
@@ -69,31 +69,14 @@ internal class DatabricksStatement : SparkStatement, IHiveServer2Statement
...
-        // Used to scope CANCEL_STATEMENT timing within the statement's lifetime.
-        private readonly Stopwatch _statementLifetimeStopwatch = Stopwatch.StartNew();
-        private bool _closeStatementTelemetryEmitted;
-
-        // Populated by DatabricksCompositeReader.Dispose when it issues TCloseOperationReq
-        // to the server. Lets us report the actual close-RPC latency in the CLOSE_STATEMENT
-        // telemetry event emitted at statement Dispose, instead of a meaningless 0.
-        internal long? CloseStatementRpcLatencyMs { get; set; }
-        internal Exception? CloseStatementRpcError { get; set; }
+        internal bool? RunAsyncOverride { get; set; }
```
eric-wang-1990, line 79: Why do we need this extra flag? Who is passing this in?
Can we just rely on runAsyncInThrift?

csharp/src/Telemetry/ConnectionTelemetry.cs
```diff
@@ -645,145 +349,36 @@ internal static Proto.DriverConnectionParameters BuildDriverConnectionParams(
                     // Bare hostname, matching JDBC. Scheme is implicit (always https) and
                     // port is reported in the sibling Port field, so including them here
```
msrathore-db, line 646: The PR does not mention any telemetry related changes. Was this supposed to be a part of the PR?

msrathore-db (approved):
LGTM apart from Eric's comment

csharp/src/DatabricksStatement.cs
jadewang-db: `runAsyncInThrift` is the user's connection-level setting (`adbc.databricks.enable_run_async_in_thrift_op`, defaults to `true`) — flipping it on the connection would affect *all* their statements, and mutating it around the desc-stmt call would race with any concurrent statements on the same connection.

`RunAsyncOverride` is an `internal` per-statement mechanism — no user surface — used exclusively by `GetColumnsExtendedAsync` to flip RunAsync=false on its own internal descStmt while the user's other statements keep their configured setting. The pairing with the `STATIC ONLY` SQL modifier is what tells the warehouse to route the desc-stmt off the WLM path; the connection-level flag alone can't express "this single statement, no others."

Open to a different shape if you have one in mind — happy to refactor if e.g. you'd prefer a constructor param or a dedicated `ExecuteWithRunAsyncOff()` helper instead of the override property.

csharp/src/Telemetry/ConnectionTelemetry.cs
jadewang-db: Good catch — that change shouldn't have been in this PR. The telemetry `host_url` fix was already landed independently via #401 (squash commit `640fbbd`), but my branch was carrying a duplicate unsquashed commit (`d02dae2`) that I hadn't noticed.

Resolved by today's rebase onto current main: `d02dae2` was dropped as a no-op (its content already in main via #401), and the PR diff no longer touches `ConnectionTelemetry.cs`. Sorry for the noise.

eric-wang-1990 (approved):

PR #460, opened 2026-05-19
refactor(csharp): make ConnectionTelemetry.Create protocol-agnostic (PECO-3022)

jadewang-db:
## 🥞 Stacked PR
Use this [link](https://github.com/adbc-drivers/databricks/pull/460/files) to review incremental changes.
- [stack/PECO-3022-sea-telemetry-design](https://github.com/adbc-drivers/databricks/pull/455) [[Files changed](https://github.com/adbc-drivers/databricks/pull/455/files)] [MERGED]
  - [**stack/pr-phase1-connection-telemetry-create-refactor**](https://github.com/adbc-drivers/databricks/pull/460) [[Files changed](https://github.com/adbc-drivers/databricks/pull/460/files)]
    - [stack/pr-phase2-observer-interface](https://github.com/adbc-drivers/databricks/pull/461) [[Files changed](https://github.com/adbc-drivers/databricks/pull/461/files/2454e7cd10de8ed68bd6b1df4583a5e55a99057c..63365ea628a7d24192348a715eefe1af8b568df3)]
      - [stack/pr-phase3-databricks-statement-observer](https://github.com/adbc-drivers/databricks/pull/462) [[Files changed](https://github.com/adbc-drivers/databricks/pull/462/files/63365ea628a7d24192348a715eefe1af8b568df3..bc5e0318b86afe4be7d2beb22cc975d2c9dbe87a)]
        - [stack/pr-phase4-sea-connection-telemetry](https://github.com/adbc-drivers/databricks/pull/463) [[Files changed](https://github.com/adbc-drivers/databricks/pull/463/files/bc5e0318b86afe4be7d2beb22cc975d2c9dbe87a..44b20018e9c58c114b05be973e9d7c2e5dd6adf5)]
          - [stack/pr-phase5-sea-statement-telemetry](https://github.com/adbc-drivers/databricks/pull/464) [[Files changed](https://github.com/adbc-drivers/databricks/pull/464/files/44b20018e9c58c114b05be973e9d7c2e5dd6adf5..f7be471df206da95f9cd9368b7b8b1ab38795a07)]

---------
## Summary

Refactors `ConnectionTelemetry.Create` so it can serve both transports cleanly — the precondition for wiring SEA telemetry in later phases of this stack:

- **Replaces `TSessionHandle? sessionHandle` with `string sessionId`.** `TSessionHandle` is a Thrift-only type; SEA only has a `string _sessionId`. The Thrift call site converts at the boundary (`new Guid(SessionHandle.SessionId.Guid).ToString()`), keeping the conversion local to Thrift code.
- **Adds `DriverMode mode` parameter.** Removes the hardcoded `DriverMode.Types.Type.Thrift` from `BuildDriverConnectionParams`/`SafeBuildDriverConnectionParams` (lines 458, 642 pre-refactor). Without this, SEA records would still be labeled `DRIVER_MODE_THRIFT`.
- **Aligns the `driver_name` constant** (`ADBC Databricks Driver`) used by both transports so lumberjack dashboards see a single literal regardless of protocol.

## Files touched

- `csharp/src/Telemetry/ConnectionTelemetry.cs` — `Create` signature change, mode threading, hardcoded Thrift removed.
- `csharp/src/DatabricksConnection.cs` — Thrift caller converts `TSessionHandle → string` and passes `DriverMode.Thrift`.
- `csharp/test/Unit/Telemetry/ConnectionTelemetryCreateSignatureTests.cs` (new) — covers `Create_AcceptsStringSessionId`, `Create_ThriftMode_SetsDriverModeThrift`, `Create_SeaMode_SetsDriverModeSea`.

## Test plan

- [x] All existing telemetry unit tests pass unchanged.
- [x] Existing Thrift integration tests continue to emit `driver_connection_params.mode = DRIVER_MODE_THRIFT`.
- [x] New unit tests cover string sessionId + both DriverMode values.

## Related

- Design: #455 — `docs/designs/PECO-3022-sea-telemetry-integration-design.md` §5.2 + §10.
- Stack root: [stack/PECO-3022-sea-telemetry-design](https://github.com/adbc-drivers/databricks/pull/455).

Part of PECO-3022.

### DIFF
--- a/csharp/src/DatabricksConnection.cs
+++ b/csharp/src/DatabricksConnection.cs
@@ -732,10 +732,18 @@ namespace AdbcDrivers.Databricks
             // server-assigned session id string directly.
             //
             // Wrap the byte[] -> Guid conversion locally: `new Guid(byte[])` throws
-            // ArgumentException on a wrong-length array, and that would propagate to
-            // connection-open. Pre-refactor the same conversion was inside
-            // ConnectionTelemetry.Create's outer try/catch, so a malformed session GUID
-            // degraded to NoOp telemetry — preserve that fail-open contract.
+            // ArgumentException on a wrong-length array, and that must not propagate
+            // to connection-open.
+            //
+            // Behavior on conversion failure: sessionId stays empty, and
+            // ConnectionTelemetry.Create maps that to SessionId = null on a live
+            // TelemetrySessionContext (see ConnectionTelemetry.cs ~L133). Telemetry
+            // remains enabled — only the session-id field is unset. This is a small,
+            // deliberate behavior change from pre-refactor, where the same conversion
+            // sat inside Create's outer try/catch and a bad GUID returned the NoOp
+            // telemetry instance. Both paths keep the *connection* fail-open; the new
+            // path additionally keeps telemetry on so we still emit driver_connection_params,
+            // driver_system_configuration, and error events for the affected session.
             string sessionId = string.Empty;
             try
             {
@@ -746,8 +754,9 @@ namespace AdbcDrivers.Databricks
             }
             catch
             {
-                // Intentionally swallowed: malformed session GUID disables telemetry,
-                // not the connection.
+                // Intentionally swallowed. Leaves sessionId = string.Empty, which
+                // Create maps to SessionId = null on a live ConnectionTelemetry.
+                // See block comment above for the deliberate behavior choice.
             }
 
             _telemetry = Telemetry.ConnectionTelemetry.Create(
--- a/csharp/src/DatabricksConnection.cs
+++ b/csharp/src/DatabricksConnection.cs
@@ -727,12 +727,20 @@ namespace AdbcDrivers.Databricks
         /// </summary>
         private void InitializeTelemetry(Activity? activity = null)
         {
+            // Convert TSessionHandle -> string at the transport boundary so
+            // ConnectionTelemetry.Create stays transport-agnostic. SEA will pass its
+            // server-assigned session id string directly.
+            string sessionId = SessionHandle?.SessionId?.Guid != null
+                ? new Guid(SessionHandle.SessionId.Guid).ToString()
+                : string.Empty;
+
             _telemetry = Telemetry.ConnectionTelemetry.Create(
                 properties: Properties,
                 host: GetHost(),
                 assemblyVersion: s_assemblyVersion,
                 oauthTokenProvider: _oauthTokenProvider,
-                sessionHandle: SessionHandle,
+                sessionId: sessionId,
+                mode: Telemetry.Proto.DriverMode.Types.Type.Thrift,
                 enableDirectResults: _enableDirectResults,
                 useDescTableExtended: _useDescTableExtended,
                 connectTimeoutMilliseconds: ConnectTimeoutMilliseconds,
--- a/csharp/src/DatabricksConnection.cs
+++ b/csharp/src/DatabricksConnection.cs
@@ -730,9 +730,25 @@ namespace AdbcDrivers.Databricks
             // Convert TSessionHandle -> string at the transport boundary so
             // ConnectionTelemetry.Create stays transport-agnostic. SEA will pass its
             // server-assigned session id string directly.
-            string sessionId = SessionHandle?.SessionId?.Guid != null
-                ? new Guid(SessionHandle.SessionId.Guid).ToString()
-                : string.Empty;
+            //
+            // Wrap the byte[] -> Guid conversion locally: `new Guid(byte[])` throws
+            // ArgumentException on a wrong-length array, and that would propagate to
+            // connection-open. Pre-refactor the same conversion was inside
+            // ConnectionTelemetry.Create's outer try/catch, so a malformed session GUID
+            // degraded to NoOp telemetry — preserve that fail-open contract.
+            string sessionId = string.Empty;
+            try
+            {
+                if (SessionHandle?.SessionId?.Guid != null)
+                {
+                    sessionId = new Guid(SessionHandle.SessionId.Guid).ToString();
+                }
+            }
+            catch
+            {
+                // Intentionally swallowed: malformed session GUID disables telemetry,
+                // not the connection.
+            }
 
             _telemetry = Telemetry.ConnectionTelemetry.Create(
                 properties: Properties,
--- a/csharp/test/Unit/Telemetry/ConnectionTelemetryCreateSignatureTests.cs
+++ b/csharp/test/Unit/Telemetry/ConnectionTelemetryCreateSignatureTests.cs
@@ -1,0 +1,173 @@
+/*
+* Copyright (c) 2025 ADBC Drivers Contributors
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using AdbcDrivers.Databricks.Telemetry;
+using AdbcDrivers.HiveServer2.Spark;
+using DriverModeType = AdbcDrivers.Databricks.Telemetry.Proto.DriverMode.Types.Type;
+using Xunit;
+
+namespace AdbcDrivers.Databricks.Tests.Unit.Telemetry
+{
+    /// <summary>
+    /// Tests for PECO-3022 (TELEM/SEA T1): <see cref="ConnectionTelemetry.Create"/>
+    /// is now transport-agnostic — it takes a <c>string sessionId</c> (converted at the
+    /// caller's boundary) and a <c>DriverMode.Types.Type mode</c> threaded through to
+    /// <c>driver_connection_params.mode</c>. The two formerly hardcoded
+    /// <c>DriverMode.Types.Type.Thrift</c> literals in <c>BuildDriverConnectionParams</c>
+    /// and the fallback in <c>SafeBuildDriverConnectionParams</c> are gone; the mode is
+    /// always the value supplied by the caller (THRIFT for Thrift, SEA for the upcoming
+    /// SEA transport).
+    /// </summary>
+    public class ConnectionTelemetryCreateSignatureTests
+    {
+        private const string AssemblyVersion = "1.2.3-test";
+        private const int DefaultTimeoutMs = 30_000;
+
+        private static IReadOnlyDictionary<string, string> TelemetryEnabledProperties() =>
+            new Dictionary<string, string>
+            {
+                { TelemetryConfiguration.PropertyKeyEnabled, "true" },
+                { SparkParameters.AuthType, SparkAuthTypeConstants.Token },
+                { SparkParameters.Token, "dapi-redacted" },
+                { SparkParameters.Path, "/sql/1.0/warehouses/abc123" },
+            };
+
+        // Tests share a TelemetryClient cache keyed by host (TelemetryClientManager
+        // singleton). Use distinct hosts per test to keep them isolated.
+        [Fact]
+        public async Task Create_AcceptsStringSessionId()
+        {
+            // Regression: the original signature took TSessionHandle?, forcing the
+            // (Thrift) caller to leak its transport handle through telemetry. The new
+            // signature accepts the already-stringified id so the SEA caller can pass
+            // its server-assigned id without inventing a fake TSessionHandle.
+            const string Host = "create-string-sid.databricks.com";
+            const string SessionId = "9e6a3f88-1234-4321-abcd-deadbeefcafe";
+
+            IConnectionTelemetry telemetry = ConnectionTelemetry.Create(
+                properties: TelemetryEnabledProperties(),
+                host: Host,
+                assemblyVersion: AssemblyVersion,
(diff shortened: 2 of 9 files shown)


csharp/src/DatabricksConnection.cs
```diff
@@ -723,12 +723,20 @@ internal IConnectionTelemetry TelemetryForTesting
         /// </summary>
         private void InitializeTelemetry(Activity? activity = null)
         {
+            // Convert TSessionHandle -> string at the transport boundary so
+            // ConnectionTelemetry.Create stays transport-agnostic. SEA will pass its
+            // server-assigned session id string directly.
+            string sessionId = SessionHandle?.SessionId?.Guid != null
+                ? new Guid(SessionHandle.SessionId.Guid).ToString()
```
jadewang-db: **[High] Fail-open contract regression.**

The `new Guid(SessionHandle.SessionId.Guid)` call now lives *outside* any try/catch. Pre-refactor, this exact conversion was on line ~120 *inside* `ConnectionTelemetry.Create`'s outer try/catch (the catch at `ConnectionTelemetry.cs:150`), so any throw degraded to `NoOpConnectionTelemetry.Instance` and the connection still opened.

`new Guid(byte[])` throws `ArgumentException` when the array is not exactly 16 bytes. The null-guard above only protects against `Guid == null`, not against a wrong-length array. While the Thrift server should always send a 16-byte payload, the explicit `Never throws` contract documented two lines above (`ConnectionTelemetry.cs:62`) is now reachable from `InitializeTelemetry` callers — a malformed session GUID would fail connection-open, not just disable telemetry.

Fix: wrap the conversion locally:

```csharp
string sessionId = string.Empty;
try
{
    if (SessionHandle?.SessionId?.Guid != null)
    {
        sessionId = new Guid(SessionHandle.SessionId.Guid).ToString();
    }
}
catch
{
    // Preserve fail-open: malformed session GUID disables telemetry, not the connection.
}
```

Alternatively, move the conversion back inside `Create` by accepting `byte[]? rawSessionGuid` — but then SEA's caller has to construct a fake byte array, which defeats the whole refactor. The local try/catch is the cheaper fix.

csharp/test/Unit/Telemetry/ConnectionTelemetryDriverNameTests.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        [Fact]
+        public void SafeBuildSystemConfiguration_ReturnsCanonicalDriverName_FallbackPath()
+        {
+            // Fallback path: even if BuildSystemConfiguration throws and the safe wrapper
+            // returns a best-effort proto, the driver_name field must still pin to the
+            // canonical literal so SEA/Thrift parity holds on partial-init failures.
+            // We exercise the wrapper directly with a degenerate assembly version — any
+            // throw inside the wrapper falls back to the literal-populated proto.
+            var config = ConnectionTelemetry.SafeBuildSystemConfiguration(string.Empty, activity: null);
```
jadewang-db: **[Medium] Test name promises a fallback that isn't actually exercised.**

The test passes `assemblyVersion: string.Empty` expecting that to trip the catch block in `SafeBuildSystemConfiguration`. But `CreateSystemConfiguration` at `ConnectionTelemetry.cs:512` does:

```csharp
DriverVersion = assemblyVersion ?? string.Empty,
```

Empty string is coalesced, not validated — it flows through happily. So `_FallbackPath` is actually exercising the same code path as `_HappyPath` above it, just with a different `DriverVersion` value. The catch block in `SafeBuildSystemConfiguration` is never reached.

Two ways to fix:

1. **Rename to reflect reality:** drop `_FallbackPath` — there's nothing in the catch worth covering separately from the happy path.
2. **Actually trigger the fallback** — harder with static methods, but `BuildSystemConfiguration` can throw via the `TrySet(() => config.OsArch = RuntimeInformation.OSArchitecture.ToString());` chain on some constrained runtimes. A cleaner approach would be to add an `internal` overload that takes a fault-injection delegate, or to assert the literal directly on the `catch` fallback proto using reflection-based invocation of the catch path.

Option 1 is fine — the *intent* (pin the literal) is already covered by `CanonicalConstant_HasExpectedLiteralValue`. The fallback path can be tested separately if/when a real fault-injection seam is added.

csharp/test/Unit/Telemetry/ConnectionTelemetryCreateSignatureTests.cs
```diff
@@ -0,0 +1,173 @@
...
+                Assert.Equal(DriverModeType.Sea, telemetry.Session.DriverConnectionParams!.Mode);
+            }
+            finally
+            {
+                await telemetry.DisposeAsync();
+            }
+        }
+
+        [Fact]
+        public void Create_ThrowingHttpClient_ReturnsNullConnectionTelemetry()
```
jadewang-db: **[Low] Test name says `ReturnsNull...` but the method returns the `NoOp` singleton, not null.**

The assertion two lines down — `Assert.Same(NoOpConnectionTelemetry.Instance, telemetry);` — confirms the actual contract: a NoOp instance, never null. Suggest renaming to `Create_ThrowingHttpClient_ReturnsNoOpConnectionTelemetry`. The current name will confuse the next reader who greps for `Null` thinking we have a null-return path.

csharp/test/Unit/Telemetry/ConnectionTelemetryCreateSignatureTests.cs
```diff
@@ -0,0 +1,173 @@
...
+                mode: DriverModeType.Thrift,
+                enableDirectResults: true,
+                useDescTableExtended: false,
+                connectTimeoutMilliseconds: DefaultTimeoutMs,
+                activity: null);
+
+            Assert.Same(NoOpConnectionTelemetry.Instance, telemetry);
+            Assert.Null(telemetry.Session);
+        }
+    }
```
jadewang-db, line 215: **[Low] Brittle test — depends on the *current* implementation of `HttpClientFactory` / `TelemetryClientManager` throwing on empty host.**

The comment is honest about the mechanism (`new Uri("https://")` throws), but that's two layers down from the unit under test. If `HttpClientFactory.CreateTelemetryHttpClient` or `TelemetryClientManager.GetOrCreateClient` later gains an early defensive check that returns gracefully on empty host, this test would still pass — but it would no longer be testing what its name suggests. It'd be a silent non-test.

Suggestion (non-blocking): add a one-line `Assert.True(string.IsNullOrEmpty(host), "This test relies on empty host tripping the inner URI/client construction")` near the top, or pull the fault-injection seam up (an internal overload of `Create` that accepts the `HttpClient`) so the throw point is explicit in the test rather than implicit in transitive dependencies.

csharp/test/Unit/Telemetry/ConnectionTelemetryCreateSignatureTests.cs
```diff
@@ -0,0 +1,173 @@
...
+                Assert.NotNull(telemetry.Session);
+                Assert.Equal(SessionId, telemetry.Session!.SessionId);
+            }
+            finally
+            {
+                await telemetry.DisposeAsync();
+            }
+        }
+
+        [Fact]
```
jadewang-db, line 120: **[Low/Suggestion] Coverage gap: no test covering the `string.Empty → SessionId = null` mapping.**

`ConnectionTelemetry.cs:133` now does:

```csharp
SessionId = !string.IsNullOrEmpty(sessionId) ? sessionId : null,
```

This intentionally maps empty string to null `SessionContext.SessionId` so callers that don't yet have a session (e.g., `InitializeTelemetry` called before `OpenSession` returns a handle) get null in the telemetry payload, not an empty literal. That's good behavior — but no test pins it. Suggest adding:

```csharp
[Fact]
public async Task Create_EmptySessionId_MapsToNullInContext()
{
    var telemetry = ConnectionTelemetry.Create(
        ..., sessionId: string.Empty, mode: DriverModeType.Thrift, ...);
    try { Assert.Null(telemetry.Session?.SessionId); }
    finally { await telemetry.DisposeAsync(); }
}
```

Without this, a refactor that drops the `!string.IsNullOrEmpty` guard (and starts emitting empty-string `SessionId` to lumberjack) would not be caught by any test.

jadewang-db (commented):
## Review of PR #460 — `ConnectionTelemetry.Create` protocol-agnostic refactor

### Summary

The refactor itself is mechanically clean — `string sessionId` + `DriverMode mode` thread through correctly, hardcoded `Thrift` literals are removed, and the boundary conversion lives where it should (in the Thrift caller). The XML docs added to `Create` explain the *why* of both parameters well.

One **High** correctness issue: moving the `TSessionHandle → string Guid` conversion out of `ConnectionTelemetry.Create`'s outer try/catch and into `InitializeTelemetry` opens a path where a malformed `SessionId.Guid` byte array would propagate to connection-open instead of degrading to NoOp telemetry. Details inline.

A few **Low/Medium** test-quality nits — including a test whose name promises a fallback path it doesn't actually exercise, and a misleading test name (`ReturnsNullConnectionTelemetry` vs. the singleton). Also inline.

### Positives worth keeping

- Per-test distinct hosts to side-step the `TelemetryClientManager` per-host singleton cache — exactly the kind of subtle isolation note future readers benefit from.
- Both `DriverMode` enum values (Thrift + Sea) are covered, including the new SEA path that doesn't even have a caller in this PR yet — that's the right place to pin behavior.
- Comments reference the prior line numbers (`ConnectionTelemetry.cs:642`) so reviewers can verify the literal really moved. Nice forensic detail.

### Scope / commit hygiene

Commit `d3190ae` titled `[gap-fill][gap-1] Align driver_name string across SEA and Thrift` only adds the regression tests — the actual alignment was done in an earlier commit on `main`. Suggest retitling to `add regression tests pinning driver_name literal` so the diff matches the title; otherwise a future bisect on "when did the alignment land" will land here and find no source change.

Overall: **approve with one High issue to address before merge.**

csharp/src/DatabricksConnection.cs
jadewang-db: Done — wrapped the byte[] -> Guid conversion in a local try/catch (commit 9b31881). A malformed session GUID now degrades to `sessionId = string.Empty` (which Create maps to `SessionId = null`), preserving the pre-refactor fail-open guarantee.

csharp/test/Unit/Telemetry/ConnectionTelemetryDriverNameTests.cs
jadewang-db: Done — went with option 1: removed the misleading `_FallbackPath` test and consolidated the happy path into a single `SafeBuildSystemConfiguration_ReturnsCanonicalDriverName` (commit 9b31881). Added a comment explaining why we don't separately cover the catch block: there's no fault-injection seam for the static helper, and `CanonicalConstant_HasExpectedLiteralValue` already pins the literal that both the happy-path and fallback branches reference.

csharp/test/Unit/Telemetry/ConnectionTelemetryCreateSignatureTests.cs
jadewang-db: Done — renamed to `Create_ThrowingHttpClient_ReturnsNoOpConnectionTelemetry` (commit 9b31881).

csharp/test/Unit/Telemetry/ConnectionTelemetryCreateSignatureTests.cs
jadewang-db: Done — added an `ASSUMPTION:` paragraph to the test's comment block that documents the implicit dependency on `HttpClientFactory` / `TelemetryClientManager` throwing on empty host, with explicit guidance for the future case where they get defensive handling and this test silently becomes a non-test (commit 9b31881). Left the underlying mechanism as-is for now — a real fault-injection seam can wait until a second consumer needs it.

csharp/test/Unit/Telemetry/ConnectionTelemetryCreateSignatureTests.cs
jadewang-db: Done — added `Create_EmptySessionId_MapsToNullInContext` (commit 9b31881). It calls `Create` with `sessionId: string.Empty` and asserts `telemetry.Session.SessionId` is null, pinning the mapping at `ConnectionTelemetry.cs:133`. Test passes locally (5/5 in `ConnectionTelemetryCreateSignatureTests`).

eric-wang-1990:
### Code review

Found 1 issue:

1. The new `InitializeTelemetry` catch block claims to preserve the pre-refactor fail-open contract, but it does not. The comments state the malformed-GUID path "degraded to NoOp telemetry — preserve that fail-open contract" and that a "malformed session GUID disables telemetry." Pre-refactor, `new Guid(...)` ran inside `ConnectionTelemetry.Create`'s outer `try { ... } catch (Exception ex) { return NoOpConnectionTelemetry.Instance; }`, so a bad GUID returned NoOp telemetry (disabled). Post-refactor, the local `catch` leaves `sessionId = string.Empty`, and `Create` then maps that to `SessionId = null` and returns a live `ConnectionTelemetry` — telemetry stays enabled, just without a session id. The behavior change may be acceptable (arguably better), but the comments assert the opposite of what the code does and will mislead future maintainers.

https://github.com/adbc-drivers/databricks/blob/2454e7cd10de8ed68bd6b1df4583a5e55a99057c/csharp/src/DatabricksConnection.cs#L735-L751

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

eric-wang-1990 (approved):

jadewang-db:
Good catch — the comments described pre-refactor behavior, not what the code actually does. Fixed in the next commit: rewrote the block comment to spell out that on conversion failure `sessionId` stays empty, `Create` maps that to `SessionId = null` on a live `ConnectionTelemetry`, and telemetry remains enabled (we still emit driver_connection_params / system_config / error events for the affected session). Behavior left as-is per your "arguably better" call. Only the comments changed.

PR #461, opened 2026-05-19
feat(csharp): add IStatementOperationObserver and fail-open implementations (PECO-3022)

jadewang-db:
## 🥞 Stacked PR
Use this [link](https://github.com/adbc-drivers/databricks/pull/461/files/2454e7cd10de8ed68bd6b1df4583a5e55a99057c..63365ea628a7d24192348a715eefe1af8b568df3) to review incremental changes.
- [stack/PECO-3022-sea-telemetry-design](https://github.com/adbc-drivers/databricks/pull/455) [[Files changed](https://github.com/adbc-drivers/databricks/pull/455/files)] [MERGED]
  - [stack/pr-phase1-connection-telemetry-create-refactor](https://github.com/adbc-drivers/databricks/pull/460) [[Files changed](https://github.com/adbc-drivers/databricks/pull/460/files)]
    - [**stack/pr-phase2-observer-interface**](https://github.com/adbc-drivers/databricks/pull/461) [[Files changed](https://github.com/adbc-drivers/databricks/pull/461/files/2454e7cd10de8ed68bd6b1df4583a5e55a99057c..63365ea628a7d24192348a715eefe1af8b568df3)]
      - [stack/pr-phase3-databricks-statement-observer](https://github.com/adbc-drivers/databricks/pull/462) [[Files changed](https://github.com/adbc-drivers/databricks/pull/462/files/63365ea628a7d24192348a715eefe1af8b568df3..bc5e0318b86afe4be7d2beb22cc975d2c9dbe87a)]
        - [stack/pr-phase4-sea-connection-telemetry](https://github.com/adbc-drivers/databricks/pull/463) [[Files changed](https://github.com/adbc-drivers/databricks/pull/463/files/bc5e0318b86afe4be7d2beb22cc975d2c9dbe87a..44b20018e9c58c114b05be973e9d7c2e5dd6adf5)]
          - [stack/pr-phase5-sea-statement-telemetry](https://github.com/adbc-drivers/databricks/pull/464) [[Files changed](https://github.com/adbc-drivers/databricks/pull/464/files/44b20018e9c58c114b05be973e9d7c2e5dd6adf5..f7be471df206da95f9cd9368b7b8b1ab38795a07)]

---------
## Summary

Adds the `IStatementOperationObserver` interface that decouples statement classes from telemetry's data model, plus three implementations. **Pure new code with no consumers yet** — wired up in subsequent stack phases.

- **`IStatementOperationObserver`** — 8 methods shaped around the statement's lifecycle (`OnExecuteStarted`, `OnExecuteSucceeded`, `OnPollCompleted`, `OnFirstBatchReady`, `OnConsumed`, `OnChunksDownloaded`, `OnError`, `OnFinalized`). Contract is documented as fail-open: implementations **MUST NOT throw**, must be thread-safe, and `OnFinalized` is terminal + idempotent. The interface speaks the statement's language, not the telemetry proto's — future tracing/audit observers can plug in without touching statement code.
- **`TelemetryObserver`** — translates observer calls to mutations on a `StatementTelemetryContext`. On `OnFinalized` builds the `OssSqlDriverTelemetryLog` and enqueues to the session's `TelemetryClient`. Uses a private `Safe(Action)` helper to concentrate the try/catch in exactly one place; method bodies are clean one-liners. Idempotency via `Interlocked.CompareExchange` on an `_emitted` flag.
- **`NullObserver`** — singleton no-op. Used as the default field value so statement code never needs null-checks at callsites.
- **`SafeObserver`** — belt-and-suspenders decorator that wraps any inner observer with per-method try/catch. Optional; first-party observers already honor the contract internally.

## Files touched (all new)

- `csharp/src/Telemetry/IStatementOperationObserver.cs`
- `csharp/src/Telemetry/TelemetryObserver.cs`
- `csharp/src/Telemetry/NullObserver.cs`
- `csharp/src/Telemetry/SafeObserver.cs`
- `csharp/test/Unit/Telemetry/TelemetryObserverTests.cs`
- `csharp/test/Unit/Telemetry/NullObserverTests.cs`
- `csharp/test/Unit/Telemetry/SafeObserverTests.cs`

## Test plan

- [x] `NullObserver_AllMethods_AreNoOps`, `NullObserver_IsSingleton`.
- [x] `TelemetryObserver_OnExecuteStarted_PopulatesContext`, `_OnExecuteSucceeded_RecordsStatementId`, `_OnFinalized_EnqueuesExactlyOnce`, `_OnFinalized_CalledTwice_EnqueuesOnce`, `_OnError_RecordsErrorAndFinalizes`, `_AllMethods_NeverThrow_WhenContextCorrupted`, `_OnChunksDownloaded_MergesIntoChunkDetails`.
- [x] `SafeObserver_PropagatesNormalCallsToInner`, `_SwallowsExceptionsFromInner_LogsAtTrace`.

## Related

- Design: #455 — §5.1 (interface contract), §12 (fail-open implementation pattern).
- Builds on #460 (no functional dependency; co-stacks for clean review).

Part of PECO-3022.


jadewang-db (commented):
Code review of phase 2 (observer interface + 3 implementations + tests).

**Summary.** Clean, well-documented, fail-open design. `Interlocked.CompareExchange` correctly guards exactly-once finalize; tests cover the contract dimensions that matter (concurrency, null inputs, throwing dependencies, no-ambient-Activity). Pure new code with no consumers yet, so blast radius is contained.

**Findings, all medium-or-lower** — none are blockers, but the comment/code mismatch in `TelemetryObserver.OnError` is worth fixing in this PR because it will mislead the next reader.

1. `TelemetryObserver.OnError` — comment contradicts the code.
2. `TelemetryObserver.Safe` — uses `ex.Message` directly; should use `SafeMessage(ex)` for parity with `OnError` and `SafeObserver`.
3. `Safe()` / `SafeMessage()` duplicated across `TelemetryObserver` and `SafeObserver` with subtly different suppression-event payloads — converge or extract.
4. Contract says "any further calls" after `OnFinalized` are no-ops; today only further `OnFinalized` is gated, other methods still mutate `_ctx`. Inert in current callers, but a strict reading of the contract would gate all methods.

Nothing critical. Approving in spirit once #1 is addressed.

csharp/src/Telemetry/TelemetryObserver.cs
```diff
@@ -0,0 +1,256 @@
...
+                if (ex == null)
+                {
+                    return;
+                }
+                _ctx.HasError = true;
+                // GetType().Name is always safe; do not capture ex.Message here unless we
+                // have explicit consent: the proto's DriverErrorInfo.error_message field is
+                // pending LPP review (see ConnectionTelemetry.EmitOperationTelemetry).
+                _ctx.ErrorName = ex.GetType().Name;
+                _ctx.ErrorMessage = SafeMessage(ex);
```
jadewang-db: **[Medium] Comment contradicts the code.**

Lines 164–166 say "do not capture `ex.Message` here unless we have explicit consent" — but line 168 then captures it: `_ctx.ErrorMessage = SafeMessage(ex);`.

The in-memory capture matches existing behavior in `ConnectionTelemetry.cs:207`, and `BuildTelemetryLog` does **not** serialize `ErrorMessage` into the proto today, so there is no actual leak. But the comment is misleading and the future risk is real: if anyone wires `ErrorMessage` into `DriverErrorInfo.error_message` later (once LPP review lands), they will reasonably assume the capture site already considered the privacy implications — when in fact the comment told them the opposite.

Pick one: either (a) drop the `_ctx.ErrorMessage = SafeMessage(ex);` line and rely on `ErrorName` only, matching what the comment says, or (b) update the comment to reflect that `.Message` *is* captured in-memory today, the proto field is gated separately in `BuildTelemetryLog`, and `SafeMessage` is used to neutralize exceptions that throw from their `.Message` property. Option (b) is consistent with `ConnectionTelemetry` and is probably what you want.

csharp/src/Telemetry/TelemetryObserver.cs
```diff
@@ -0,0 +1,256 @@
...
+            }
+            catch (Exception ex)
+            {
+                try
+                {
+                    Activity.Current?.AddEvent(new ActivityEvent("telemetry.observer.suppressed",
+                        tags: new ActivityTagsCollection
+                        {
+                            { "error.type", ex.GetType().Name },
+                            { "error.message", ex.Message }
```
jadewang-db: **[Medium] `ex.Message` here, but `SafeMessage(ex)` in `OnError`** (and in `SafeObserver`).

The rest of this file — and `SafeObserver.Safe` — explicitly guard against exceptions whose `.Message` property throws. This site doesn't. The outer `catch {}` on line 234 saves us from a crash, but it also swallows the suppression event for that specific failure mode, so we lose telemetry visibility into observer failures exactly when an upstream exception is misbehaving.

Use `SafeMessage(ex)` for consistency.

csharp/src/Telemetry/SafeObserver.cs
```diff
@@ -0,0 +1,147 @@
...
+            catch (Exception ex)
+            {
+                try
+                {
+                    Activity.Current?.AddEvent(new ActivityEvent("telemetry.observer.suppressed",
+                        tags: new ActivityTagsCollection
+                        {
+                            { "error.type", ex.GetType().Name },
+                            { "error.message", SafeMessage(ex) },
+                            { "observer.suppressed.source", "SafeObserver" },
```
jadewang-db: **[Low] Diverging suppression-event payloads between the two `Safe()` helpers.**

`SafeObserver.Safe` tags the suppression event with `observer.suppressed.source = "SafeObserver"` (line 122). `TelemetryObserver.Safe` (`TelemetryObserver.cs:227`) emits the same event name (`telemetry.observer.suppressed`) but without a source tag.

This matters in two ways:
- A trace consumer can't tell which layer suppressed an exception when both decorators are in play (e.g. a third-party observer wrapped in `SafeObserver` wrapped over `TelemetryObserver`).
- These two helpers — plus `SafeMessage` — are essentially duplicated. Converging them into a single internal helper (e.g. `TelemetryObserverSafety.Run(Action, string source)`) would make the inconsistency impossible by construction. Optional, but worth doing while the surface is small.

csharp/src/Telemetry/IStatementOperationObserver.cs
```diff
@@ -0,0 +1,114 @@
...
+    /// <b>Thread-safety:</b> Methods on this interface may be invoked from any thread.
+    /// Implementations <b>MUST be thread-safe</b>. In practice the calls happen from the
+    /// statement's executing task and the reader's disposal thread, which may differ.
+    /// </para>
+    /// <para>
+    /// <b>Terminal call:</b> <see cref="OnFinalized"/> is the terminal call. After it has
+    /// been invoked once, the observer's record is considered complete and any further
+    /// calls — including additional <see cref="OnFinalized"/> calls — <b>MUST be no-ops
+    /// (idempotent)</b>. This protects against the common case where both an error path
+    /// and a dispose path attempt to finalize the same statement.
```
jadewang-db, line 47: **[Low] Contract vs. implementation: post-`OnFinalized` non-`OnFinalized` calls.**

This contract says: "any further calls — including additional `OnFinalized` calls — MUST be no-ops (idempotent)." That language covers *all* methods, not just `OnFinalized`.

`TelemetryObserver` only gates further `OnFinalized` via `_emitted`. After finalize, a stray `OnPollCompleted` / `OnError` / `OnChunksDownloaded` will still mutate `_ctx`. The mutations are inert today because the log has already been built and no one reads the context post-finalize — but a strict reading of the contract would have all observer methods check `_emitted` first and return early.

Two equally reasonable resolutions:
- Loosen the interface doc: "`OnFinalized` is the only terminal call that must be idempotent; other methods after finalize are silently dropped but their effect on observer state is implementation-defined."
- Tighten the implementation: add an `if (Volatile.Read(ref _emitted) != 0) return;` short-circuit at the top of every `Safe(...)` body in `TelemetryObserver`.

The contract-shipped-as-written is what later reviewers will hold us to, so picking one of these and aligning is worth doing now while there are no external consumers.

csharp/test/Unit/Telemetry/TelemetryObserverTests.cs
```diff
@@ -0,0 +1,367 @@
...
+                observer.OnError(new InvalidOperationException("boom"));
+                observer.OnFinalized();
+                observer.OnFinalized();
+            });
+
+            Assert.Null(captured);
+
+            // The throwing client must have been invoked exactly once (idempotent finalize)
+            // and the observer must have swallowed its exception.
+            Assert.Equal(1, throwing.EnqueueCallCount);
```
jadewang-db, line 268: **[Nit] Add a test for the post-finalize behavior** that pins down whichever interpretation you pick in the contract-vs-impl issue (see comment on `IStatementOperationObserver.cs:47`).

E.g., something like `OnFinalized_ThenLifecycleCalls_AreNoOps` that asserts later `OnPollCompleted` / `OnError` calls do *not* enqueue a second log and do not affect any observable post-finalize state. Today this is implicitly true (because the second `Enqueue` is gated by `_emitted`), but having the assertion in the test suite locks it in.

Low priority.

csharp/src/Telemetry/TelemetryObserver.cs
jadewang-db: ✅ Done — went with option (b): updated the comment at `TelemetryObserver.cs:165-170` to accurately describe what the code does. The new comment notes that `.Message` is captured in-memory only, that `BuildTelemetryLog` does not serialize `ErrorMessage` into the proto today (LPP review pending), and that `SafeMessage` guards against `.Message` throwing. Matches the existing pattern in `ConnectionTelemetry.cs:207`.

csharp/src/Telemetry/TelemetryObserver.cs
jadewang-db: ✅ Done — `TelemetryObserver.Safe()` now uses `SafeMessage(ex)` instead of `ex.Message` directly. Suppression telemetry is now preserved even when the inner exception has a throwing `.Message` property.

csharp/src/Telemetry/SafeObserver.cs
jadewang-db: ✅ Done — chose to converge rather than extract a shared helper (lighter for two small classes). Added `observer.suppressed.source = "TelemetryObserver"` to `TelemetryObserver.Safe()` so a trace consumer can tell the two suppression layers apart when both decorators are in play.

csharp/src/Telemetry/IStatementOperationObserver.cs
jadewang-db: ✅ Done — tightened the implementation (the safer interpretation per the review comment). Added a private `IsFinalized()` helper (`Volatile.Read(ref _emitted) != 0`) and a gate at the top of every non-finalize lifecycle method body in `TelemetryObserver`. After `OnFinalized()` fires, all other lifecycle methods become true no-ops with no further mutation of `_ctx`.

csharp/test/Unit/Telemetry/TelemetryObserverTests.cs
jadewang-db: ✅ Done — added `TelemetryObserver_OnFinalized_ThenLifecycleCalls_AreNoOps`. Captures the context fingerprint right after `OnFinalized`, then calls every non-terminal lifecycle method with values that would visibly mutate state if the new `IsFinalized()` gate were missing, and asserts both the enqueue count and every captured field are unchanged. Locks in the contract interpretation chosen in the previous fix.

eric-wang-1990:
### Code review

Found 1 issue:

1. The observer interface contradicts this PR's own design doc, and regresses a change that was explicitly agreed in the design review (PR #455). The interface ships `OnExecuteSucceeded(string statementId, ExecutionResultFormat resultFormat)` and `OnPollCompleted(int count, long latencyMs)`:

https://github.com/adbc-drivers/databricks/blob/54e44022586e28acccddbbdfbb6ddc4e7805116c/csharp/src/Telemetry/IStatementOperationObserver.cs#L69-L79

But the design doc updated in this same PR specifies `OnStatementSubmitted(string statementId)` (no `resultFormat`) and `OnPollCompleted(int count, long latencyMs, ExecutionResult.Types.Format resultFormat)` (`resultFormat` moved to the terminal poll), with the rationale that SEA's execute returns `PENDING`/`RUNNING` (accepted, not succeeded) and the result-format manifest is null until polling reaches `SUCCEEDED`:

https://github.com/adbc-drivers/databricks/blob/54e44022586e28acccddbbdfbb6ddc4e7805116c/docs/designs/PECO-3022-sea-telemetry-integration-design.md#L234-L248

This is the exact rename + signature change raised by @eric-wang-1990 on #455 and agreed by the author ("Agreed on both points — renamed and moved format resolution"). As shipped, the SEA wiring in later phases cannot populate `ResultFormat` correctly: it isn't resolvable at submit time, and `OnPollCompleted` (where it becomes available) has no parameter to carry it. `TelemetryObserver.OnExecuteSucceeded` captures `resultFormat` at submit time, which is the behavior the design doc rejected.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

csharp/src/Telemetry/TelemetryObserver.cs
```diff
@@ -0,0 +1,272 @@
...
+                client.Enqueue(frontendLog);
+            });
+        }
+
+        /// <summary>
+        /// Concentrates the fail-open try/catch in one place. Any exception is suppressed
+        /// and surfaced as an OpenTelemetry activity event so it is still observable in
+        /// traces without affecting the caller's control flow.
+        /// </summary>
+        private static void Safe(Action action)
```
eric-wang-1990: Why duplicate `Safe` / `SafeMessage` here when `SafeObserver` already exists for the same purpose? They're byte-identical apart from the source tag and will drift over time. Pick one direction:

- Drop `SafeObserver` and keep this internal helper (third-party observers must honor the contract themselves), OR
- Always wrap (`new SafeObserver(new TelemetryObserver(...))`) and delete this internal helper.

Right now we have both, neither wraps the other, and the duplication is silent.

csharp/src/Telemetry/TelemetryObserver.cs
```diff
@@ -0,0 +1,272 @@
...
+        {
+            try
+            {
+                action();
+            }
+            catch (Exception ex)
+            {
+                try
+                {
+                    Activity.Current?.AddEvent(new ActivityEvent("telemetry.observer.suppressed",
```
eric-wang-1990, line 219: What is `Safe()` actually guarding against here? The lifecycle methods (`OnExecuteStarted` etc.) only do plain field assignments — they can't realistically throw. `Activity.Current?.AddEvent(...)` is null-guarded and `ActivityTagsCollection` accepts null values, so this inner try/catch is protecting code that doesn't fail.

The genuine risk surface is `BuildTelemetryLog()` + `client.Enqueue(...)` in `OnFinalized`. Consider scoping the try/catch to those two calls only, rather than wrapping every method (and double-nesting around `AddEvent`).

csharp/src/Telemetry/TelemetryObserver.cs
jadewang-db: Fair point. Picked option A: dropped `SafeObserver.cs` and `SafeObserverTests.cs` entirely. The original framing in the design (SafeObserver as "optional, for future third-party observers") was the weaker of the two paths — the duplication outweighed the speculative future use. Third-party observers must honor the contract themselves, same as `TelemetryObserver` and `NullObserver` do.

Also addressed the design doc + sprint plan references so we are not left with dangling pointers.

csharp/src/Telemetry/TelemetryObserver.cs
jadewang-db: You are right, the wrap was paranoia over plain field assignments. Removed `Safe(Action)` entirely and scoped the `try/catch` to the genuine risk surface — `BuildTelemetryLog()` + `client.Enqueue(...)` inside `OnFinalized`. Lifecycle methods now read as straight scalar writes with an `IsFinalized()` short-circuit at the top.

`SafeMessage` stays because `OnError` still needs to neutralize the rare exception-with-throwing-`.Message` case before storing it on the context.

PR #465, opened 2026-05-20
feat(csharp): honor cloudfetch.lz4.enabled on SEA path (PECO-3056)

eric-wang-1990:
## What's Changed

Wires `adbc.databricks.cloudfetch.lz4.enabled` through `StatementExecutionConnection` on the SEA path. Previously this param was silently ignored on SEA.

- **`cloudfetch.lz4.enabled=false`** clears `result_compression` on the wire (null → server treats as no compression). Mirrors JDBC's `CompressionCodec.NONE` branch in `DatabricksSdkClient`.
- **`cloudfetch.lz4.enabled=true`** (default) sets `result_compression=LZ4_FRAME`. The driver only implements LZ4 decompression; there is no other valid non-null value.

**`cloudfetch.enabled=false` is not wired on SEA.** Disabling CloudFetch on SEA requires the server to support `ARROW_STREAM` results with `INLINE` disposition. That server-side capability is not yet available, so the param is accepted but silently ignored on SEA for now. Support will be added in a follow-on once the server side lands.

**Test seam:** `StatementExecutionStatement.LastExecuteRequest` (internal) captures the request the driver last handed to the SEA client. E2E tests assert on `result_compression` there rather than intercepting HTTP.

## Why

PECO-3056 — https://databricks.atlassian.net/browse/PECO-3056

The M3 remaining-parameter-gap audit confirmed `cloudfetch.lz4.enabled` was ignored on SEA. This is the foundation for dependent CloudFetch sub-param tickets (PECO-3061, 3063, 3067, 3068), all of which operate with `cloudfetch.enabled=true` (the default).

## Red → Green proof

E2E tests run against the live `pecotesting` warehouse (`DATABRICKS_TEST_CONFIG_FILE` set).

**Before fix** (test seam added, no production logic — params still silently ignored):
```
[FAIL] Lz4EnabledFalse_ClearsResultCompression
  Assert.Null() Failure: Value is not null
  Expected: null
  Actual:   "LZ4_FRAME"

Failed!  - Failed: 1, Passed: 2, Skipped: 0, Total: 3
```

**After fix:**
```
Passed!  - Failed: 0, Passed: 3, Skipped: 0, Total: 3, Duration: 3 s
```

Safety run — `StatementExecutionDriverE2ETests` (11 tests, CREATE/INSERT/UPDATE/DELETE/SELECT round-trips on SEA):
```
Passed!  - Failed: 0, Passed: 11, Skipped: 0, Total: 11, Duration: 39 s
```

## Files touched

- `csharp/src/DatabricksParameters.cs` — XML doc clarifying SEA-path semantics for `UseCloudFetch` / `CanDecompressLz4`.
- `csharp/src/StatementExecution/StatementExecutionConnection.cs` — derive `result_compression` from `CanDecompressLz4` only (`LZ4_FRAME` or null).
- `csharp/src/StatementExecution/StatementExecutionStatement.cs` — internal `LastExecuteRequest` test seam.
- `csharp/test/E2E/StatementExecution/SeaCloudFetchParamsE2ETests.cs` — 3 E2E tests covering default disposition, lz4-off, and lz4-on.

## Manual verification

- [x] `dotnet build` green on `netstandard2.0` and `net8.0`
- [x] New E2E tests pass against live `pecotesting` warehouse
- [x] `StatementExecutionDriverE2ETests` (full class) passes — no regression

PECO-3056

### DIFF
--- a/csharp/src/DatabricksParameters.cs
+++ b/csharp/src/DatabricksParameters.cs
@@ -34,12 +34,32 @@ namespace AdbcDrivers.Databricks
         /// <summary>
         /// Whether to use CloudFetch for retrieving results.
         /// Default value is true if not specified.
+        ///
+        /// Honored by both protocols:
+        ///   - Thrift: controls the canUseCloudFetch flag on TFetchResultsReq.
+        ///   - SEA (REST): JDBC pairs <c>cloudfetch.enabled=false</c> with
+        ///     <c>disposition=INLINE</c> + <c>format=JSON_ARRAY</c>, because the
+        ///     SEA server rejects <c>ARROW_STREAM</c> with pure <c>INLINE</c>.
+        ///     The C# SEA driver does not yet ship a <c>JSON_ARRAY</c> reader,
+        ///     so requesting <c>cloudfetch.enabled=false</c> on the SEA path
+        ///     surfaces a clear <c>AdbcStatusCode.NotImplemented</c> error at
+        ///     connect time rather than silently ignoring the param. Once a
+        ///     JSON_ARRAY reader lands, this will flip to the JDBC behavior.
+        ///
+        /// Note: Reyden does not generate external links and will coerce results
+        /// to INLINE on the server side regardless of what the driver requests.
         /// </summary>
         public const string UseCloudFetch = "adbc.databricks.cloudfetch.enabled";
 
         /// <summary>
         /// Whether the client can decompress LZ4 compressed results.
         /// Default value is true if not specified.
+        ///
+        /// Honored by both protocols:
+        ///   - Thrift: controls canDecompressLZ4Result on TFetchResultsReq.
+        ///   - SEA (REST): when false, clears <c>result_compression</c> on the
+        ///     ExecuteStatement request (overriding any <see cref="ResultCompression"/>
+        ///     value). Mirrors JDBC's <c>CompressionCodec.NONE</c> branch.
         /// </summary>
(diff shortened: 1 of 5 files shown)


csharp/src/DatabricksParameters.cs
```diff
@@ -34,12 +34,29 @@ public class DatabricksParameters : SparkParameters
...
         /// Whether to use CloudFetch for retrieving results.
         /// Default value is true if not specified.
+        ///
+        /// Honored by both protocols:
+        ///   - Thrift: controls the canUseCloudFetch flag on TFetchResultsReq.
+        ///   - SEA (REST): not yet honored; the param is accepted but ignored.
+        ///     JDBC pairs <c>cloudfetch.enabled=false</c> with
+        ///     <c>disposition=INLINE</c> + <c>format=JSON_ARRAY</c>, but the
+        ///     C# SEA driver does not yet ship a JSON_ARRAY reader. Support
+        ///     is deferred until a JSON_ARRAY reader lands.
```
eric-wang-1990: We are waiting the Arrow_stream inline result is ready from server

eric-wang-1990: ✅ Done — updated the doc to say the blocker is server-side (waiting for ARROW_STREAM + INLINE disposition support from the server), not a missing JSON_ARRAY reader.

---

eric-wang-1990:
Superseded by #562 (default SEA result compression to LZ4, derived from cloudfetch.lz4.enabled) — what this PR aimed to do. Reopen if a gap remains.

PR #466, opened 2026-05-20
feat(csharp): honor adbc.spark.connect_timeout_ms on SEA path (PECO-3059) [SEA]

eric-wang-1990:
## What's Changed

Wires `adbc.spark.connect_timeout_ms` through `StatementExecutionConnection` so the SEA (Statement Execution REST) path bounds session bring-up with the same semantics as the Thrift / HiveServer2 path. Before this PR, the parameter was silently ignored on the SEA path: `CreateSessionAsync` ran with no cancellation token, so a hung warehouse would wait for the full `HttpClient.Timeout` (minutes) before failing.

The fix:
- Parses `adbc.spark.connect_timeout_ms` in `StatementExecutionConnection`'s constructor with Thrift-compatible validation (0 = infinite, negative = invalid, default 30000 ms — matches `HiveServer2Connection.ConnectTimeoutMillisecondsDefault`).
- Mirrors `DatabricksConnection.ValidateOptions` (lines 967-972): bumps the connect-timeout up to `temporarily_unavailable_retry_timeout * 1000` when retry is enabled, so the connect-timeout token doesn't cancel mid-retry-loop.
- Wraps `CreateSessionAsync` (and the semaphore wait) inside `OpenAsync` with a linked `CancellationTokenSource` so either the caller's token or the connect-timeout can cancel. Cancellations triggered by the connect-timeout (and not the caller's token) are translated to `TimeoutException` for exception-type parity with `HiveServer2Connection.OpenAsync`.

## Why

PECO-3059 — M3 remaining parameter gap. Connection-timeout handling on the SEA path needed parity with Thrift before the Jun 10 cutoff.

## Red → Green proof

E2E test: `AdbcDrivers.Databricks.Tests.DatabricksConnectionTest.SeaConnectionTimeout_HonorsConnectTimeoutMs` — forces `adbc.databricks.protocol=rest`, sets `adbc.spark.connect_timeout_ms=1`, and disables `temporarily_unavailable_retry` to bypass the parity-bump. The test runs against the live `pecotesting` warehouse.

**Before fix (red, with the production change reverted):**

```
[xUnit.net 00:00:00.91]     AdbcDrivers.Databricks.Tests.DatabricksConnectionTest.SeaConnectionTimeout_HonorsConnectTimeoutMs [FAIL]
  Failed AdbcDrivers.Databricks.Tests.DatabricksConnectionTest.SeaConnectionTimeout_HonorsConnectTimeoutMs [643 ms]
  Error Message:
   Assert.NotNull() Failure: Value is null
  Standard Output Messages:
 [PECO-3059] Forcing SEA protocol with ConnectTimeoutMilliseconds=1, retry disabled
 [PECO-3059] Elapsed: 640 ms. Exception: <none>

Failed!  - Failed:     1, Passed:     0, Skipped:     0, Total:     1
```

The connection succeeded in 640 ms — the 1 ms connect-timeout was silently ignored.

**After fix (green):**

```
A total of 1 test files matched the specified pattern.

Passed!  - Failed:     0, Passed:     1, Skipped:     0, Total:     1, Duration: 26 ms - AdbcDrivers.Databricks.Tests.dll (net8.0)
```

Connection fails fast (~26 ms total test duration including teardown) with a `TimeoutException` in the chain, matching the Thrift `ConnectionTimeoutTest`.

## Files touched

- `csharp/src/StatementExecution/StatementExecutionConnection.cs` — parse the timeout, apply Thrift-parity bump, gate `CreateSessionAsync` with linked CTS, translate cancellation → `TimeoutException`.
- `csharp/test/E2E/DatabricksConnectionTest.cs` — new `[SkippableFact]` E2E test asserting the SEA path fails fast under `connect_timeout_ms=1`.

## Safety

- Full `DatabricksConnectionTest` class: 93 passed, 0 failed (42 s).
- Full `StatementExecution*` E2E tests: 225 passed, 0 failed (1m 9s).
- `dotnet build` succeeds for `netstandard2.0`, `net472`, `net8.0`.

## Manual verification

- [x] E2E test (`SeaConnectionTimeout_HonorsConnectTimeoutMs`) passes against `pecotesting`
- [x] Full `DatabricksConnectionTest` class passes (no regression)
- [x] Full `StatementExecution*` test class passes (no regression)
- [x] Build green on all TFMs
- [ ] pre-commit run (skipped — sandbox blocked network access for hook env install; will run in CI)

PECO-3059

### DIFF
--- a/csharp/src/StatementExecution/StatementExecutionConnection.cs
+++ b/csharp/src/StatementExecution/StatementExecutionConnection.cs
@@ -483,21 +483,24 @@ namespace AdbcDrivers.Databricks.StatementExecution
                             var response = await _client.CreateSessionAsync(request, openToken).ConfigureAwait(false);
                             _sessionId = response.SessionId;
                         }
-                        catch (DatabricksException)
-                        {
-                            throw;
-                        }
+                        // PECO-3059: when the connect-timeout fired before the caller's token,
+                        // translate any exception (including DatabricksException-wrapped OAuth
+                        // cancellations) into TimeoutException for parity with HiveServer2Connection.
+                        // This filter must run BEFORE the unconditional DatabricksException
+                        // rethrow because OAuthClientCredentialsProvider wraps cancelled HTTP
+                        // calls as DatabricksException("Failed to acquire OAuth access token: …").
                         catch (Exception ex) when (
                             connectTimeoutCts != null &&
                             connectTimeoutCts.IsCancellationRequested &&
                             !cancellationToken.IsCancellationRequested)
                         {
-                            // PECO-3059: the connect-timeout fired before the caller's
-                            // token. Surface as TimeoutException for parity with
-                            // HiveServer2Connection.OpenAsync — same message phrasing.
                             throw new TimeoutException(
                                 "The operation timed out while attempting to open a session. Please try increasing connect timeout.",
                                 ex);
+                        }
+                        catch (DatabricksException)
+                        {
+                            throw;
                         }
                         catch (Exception ex)
                         {
--- a/csharp/src/StatementExecution/StatementExecutionConnection.cs
+++ b/csharp/src/StatementExecution/StatementExecutionConnection.cs
@@ -44,7 +44,7 @@ namespace AdbcDrivers.Databricks.StatementExecution
         private readonly string _warehouseId;
         private readonly string? _orgId;
         private string? _catalog;
-        private readonly string? _schema;
+        private string? _schema;
         private readonly HttpClient _httpClient;
         private readonly HttpClient _cloudFetchHttpClient; // Separate HttpClient without auth headers for CloudFetch downloads
         private readonly IReadOnlyDictionary<string, string> _properties;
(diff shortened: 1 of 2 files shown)


csharp/src/StatementExecution/StatementExecutionConnection.cs
```diff
@@ -324,6 +337,50 @@ private HttpClient CreateHttpClient(IReadOnlyDictionary<string, string> properti
...
         }
 
+        /// <summary>
+        /// Parses adbc.spark.connect_timeout_ms with Thrift-compatible validation and applies
+        /// the same TemporarilyUnavailableRetry adjustment as DatabricksConnection
+        /// (ValidateOptions, lines 967-972): when temporarily-unavailable retry is enabled,
+        /// the connect timeout must be at least as large as the retry budget — otherwise the
+        /// connect-timeout token would cancel mid-retry-loop before the server warms up.
+        /// </summary>
+        private static int ParseConnectTimeoutMilliseconds(IReadOnlyDictionary<string, string> properties)
```
eric-wang-1990: We should also do a similar function like DatabricksConnection.ValidateOptions where we validate all options

csharp/src/StatementExecution/StatementExecutionConnection.cs
```diff
@@ -374,50 +431,88 @@ public async Task OpenAsync(CancellationToken cancellationToken = default)
...
+                    ? CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, connectTimeoutCts.Token)
+                    : null;
+                CancellationToken openToken = linkedCts?.Token ?? cancellationToken;
+
                 try
                 {
-                    // Double-check after acquiring lock
-                    if (_sessionId == null)
+                    await _sessionLock.WaitAsync(openToken).ConfigureAwait(false);
+                    try
```
eric-wang-1990: Why you introduce one more layer of try? Can we avoid that

csharp/src/StatementExecution/StatementExecutionConnection.cs
eric-wang-1990: Done in fe4334f — introduced `ValidateOptions()` mirroring the `DatabricksConnection` pattern. All property-driven fields (result config, feature flags, tracing, identity federation, connect timeout) are now parsed and validated in that one method instead of being scattered through the constructor. `ParseConnectTimeoutMilliseconds` is gone; its logic lives inside `ValidateOptions`.

csharp/src/StatementExecution/StatementExecutionConnection.cs
eric-wang-1990: Done in fe4334f — collapsed to a single try/catch/finally using a `bool lockAcquired = false` flag. The previous outer (timeout-mapping) + inner (lock-release) try pair is now one try; the innermost try around `CreateSessionAsync` is preserved since it maps non-cancellation exceptions to `DatabricksException`.

csharp/src/StatementExecution/StatementExecutionConnection.cs
```diff
@@ -54,28 +54,35 @@ internal class StatementExecutionConnection : TracingConnection, IGetObjectsData
...
-        // Configuration for statement creation
-        private readonly string _resultDisposition;
-        private readonly string _resultFormat;
-        private readonly string? _resultCompression;
-        private readonly int _waitTimeoutSeconds;
-        private readonly int _pollingIntervalMs;
-        private readonly bool _enablePKFK;
-        private readonly bool _enableMultipleCatalogSupport;
-        private readonly bool _useDescTableExtended;
+        // Configuration for statement creation — assigned by ValidateOptions().
```
eric-wang-1990: Do not add default value to the variable definitions, they will be assigned later in ValidateOptions

csharp/src/StatementExecution/StatementExecutionConnection.cs
```diff
@@ -324,6 +292,95 @@ private HttpClient CreateHttpClient(IReadOnlyDictionary<string, string> properti
             return httpClient;
         }
 
+        /// <summary>
+        /// Parses and validates every property-driven option for this connection in one place.
+        /// Mirrors <see cref="DatabricksConnection.ValidateOptions"/> on the Thrift path so all
+        /// option-handling logic lives next to itself rather than scattered through the constructor.
+        /// </summary>
+        private void ValidateOptions()
```
eric-wang-1990: Let's call this validateProperties instead.

csharp/test/E2E/DatabricksConnectionTest.cs
```diff
@@ -171,6 +172,79 @@ public void ConnectionTimeoutTest(int? connectTimeoutMilliseconds, Type? excepti
...
+
+            OutputHelper?.WriteLine($"[PECO-3059] Elapsed: {stopwatch.ElapsedMilliseconds} ms. Exception: {caught?.GetType().FullName ?? "<none>"}");
+
+            // The connect timeout MUST fire long before the HttpClient overall timeout
+            // (which defaults to minutes on the SEA path). Allow a generous 10 s budget
+            // for TLS abort + handler unwind on a slow machine; the actual signal we
+            // care about is "fails fast", not "takes the full HTTP timeout".
+            Assert.NotNull(caught);
+            Assert.True(
+                stopwatch.ElapsedMilliseconds < 10_000,
```
eric-wang-1990: Your timeout is 1ms, why validating against 10s? That is too much diff, this should be <1s?

csharp/src/StatementExecution/StatementExecutionConnection.cs
eric-wang-1990: Done in `ac69d00` — dropped the default initializers from all property-backed fields. Used `= null!` only on the two non-nullable `string` fields (`_resultDisposition`, `_resultFormat`, `_traceParentHeaderName`) to silence CS8618 from nullable-reference-types; the bool/int/`string?` fields are now bare declarations. `ValidateProperties()` assigns each one unconditionally.

---

csharp/src/StatementExecution/StatementExecutionConnection.cs
eric-wang-1990: Done in `ac69d00` — renamed `ValidateOptions()` → `ValidateProperties()` and updated the call site and surrounding comments.

---

csharp/test/E2E/DatabricksConnectionTest.cs
eric-wang-1990: Good catch — tightened to `< 1_000` ms in `ac69d00`. Green runs complete in ~25-50 ms locally, so 1 s leaves comfortable slack for CI without being so loose that a regression to the full `HttpClient.Timeout` (minutes) would still slip through.

---

jadewang-db (approved):

PR #468, opened 2026-05-20
feat(csharp): honor adbc.databricks.apply_ssp_with_queries on SEA path (PECO-3062) [SEA]

eric-wang-1990:
## What's Changed

The SEA path silently ignored the `adbc.databricks.apply_ssp_with_queries` flag — SSPs (`adbc.databricks.ssp_*`) were always packed into `CreateSession.session_confs` regardless of the user's choice, breaking parity with the Thrift driver where the flag controls whether SSPs ride in `OpenSession.Configuration` (flag=false) or are applied via post-open `SET key=value` statements (flag=true).

This PR plumbs the flag through `StatementExecutionConnection`, gates the `CreateSessionRequest.SessionConfigs` payload on `!apply_ssp_with_queries`, and adds an `ApplyServerSidePropertiesAsync()` helper that mirrors the Thrift path's SET-statement loop. `DatabricksDatabase.Connect` now invokes the helper after the SEA `OpenAsync`, directly mirroring the existing Thrift wiring two lines below.

## Why

PECO-3062 — `adbc.databricks.apply_ssp_with_queries` M3 remaining parameter gap.

**Scope note:** the ticket envelope mentioned passing SSPs via the SEA `ExecuteStatement` request body. Verified against `StatementExecutionModels.cs` and the JDBC reference (`databricks-jdbc/src/main/java/com/databricks/jdbc/model/client/sqlexec/ExecuteStatementRequest.java`): the SEA API has no per-statement `session_configs` field. JDBC has the same constraint — and JDBC doesn't expose `apply_ssp_with_queries` at all (always uses session_confs). The achievable parity is therefore "mirror Thrift's two-mode behavior", which this PR delivers. The session retains SET state for subsequent statements, so the observable end-state matches Thrift in both modes.

## Red → Green proof

**Unit test (deterministic wire proof) — `OpenAsync_WithApplySSPWithQueries_OmitsSessionConfs`:**

Before fix:
```
[xUnit.net 00:00:00.24]     AdbcDrivers.Databricks.Tests.Unit.StatementExecution.StatementExecutionConnectionSspTests.OpenAsync_WithApplySSPWithQueries_OmitsSessionConfs [FAIL]
  Error Message:
   Assert.DoesNotContain() Failure: Sub-string found
                                ↓ (pos 51)
String: ···"","catalog":"main","session_confs":{"ansi"···
Found:  "session_confs"
Failed!  - Failed: 1, Passed: 1, Skipped: 0, Total: 2
```

After fix:
```
Passed!  - Failed: 0, Passed: 2, Skipped: 0, Total: 2, Duration: 71 ms - AdbcDrivers.Databricks.Tests.dll (net8.0)
```

**E2E tests against live pecotesting warehouse — `TestServerSidePropertyOnSeaPath`:**

```
  Passed AdbcDrivers.Databricks.Tests.ServerSidePropertyE2ETest.TestServerSidePropertyOnSeaPath(applyWithQueries: True) [2 s]
Property: ansi_mode = false
Property: enable_photon = true
Property: spark.thriftserver.arrowBasedRowSet.timestampAsString = false
Property: statement_timeout = 12345
Property: use_cached_result = false
  Passed AdbcDrivers.Databricks.Tests.ServerSidePropertyE2ETest.TestServerSidePropertyOnSeaPath(applyWithQueries: False) [1 s]
Test Run Successful.
Total tests: 2  Passed: 2
```

Full `ServerSidePropertyE2ETest` class (4 tests including the pre-existing Thrift cases) all green:
```
Test Run Successful.
Total tests: 4  Passed: 4  Total time: 5.8088 Seconds
```

## Files touched

- `csharp/src/StatementExecution/StatementExecutionConnection.cs` — parse flag; gate session_confs; add `ApplyServerSidePropertiesAsync`.
- `csharp/src/DatabricksDatabase.cs` — invoke `ApplyServerSidePropertiesAsync` on SEA path after `OpenAsync` (one-line wiring mirroring the existing Thrift call directly below).
- `csharp/test/E2E/ServerSidePropertyE2ETest.cs` — new `[SkippableTheory]` `TestServerSidePropertyOnSeaPath(applyWithQueries)` forcing `protocol=rest`.
- `csharp/test/Unit/StatementExecution/StatementExecutionConnectionSspTests.cs` — added two unit tests for the wire-level behavior.

## Manual verification

- [x] Build green on `netstandard2.0` + `net8.0` (macOS does not include `net472` in the build matrix)
- [x] All 9 unit tests in `StatementExecutionConnectionSspTests` pass
- [x] All 4 E2E tests in `ServerSidePropertyE2ETest` pass against live warehouse
- [ ] `pre-commit run --all-files` — environmental failure (sandbox blocks network for ruff/golangci installs); no C# pre-commit hook configured

PECO-3062

### DIFF
--- a/csharp/src/StatementExecution/StatementExecutionConnection.cs
+++ b/csharp/src/StatementExecution/StatementExecutionConnection.cs
@@ -63,6 +63,7 @@ namespace AdbcDrivers.Databricks.StatementExecution
         private readonly bool _enablePKFK;
         private readonly bool _enableMultipleCatalogSupport;
         private readonly bool _useDescTableExtended;
+        private readonly bool _applySSPWithQueries;
 
         // Memory pooling (shared across connection)
         private readonly Microsoft.IO.RecyclableMemoryStreamManager _recyclableMemoryStreamManager;
@@ -204,6 +205,12 @@ namespace AdbcDrivers.Databricks.StatementExecution
             _enablePKFK = PropertyHelper.GetBooleanPropertyWithValidation(properties, DatabricksParameters.EnablePKFK, true);
             _enableMultipleCatalogSupport = PropertyHelper.GetBooleanPropertyWithValidation(properties, DatabricksParameters.EnableMultipleCatalogSupport, true);
             _useDescTableExtended = PropertyHelper.GetBooleanPropertyWithValidation(properties, DatabricksParameters.UseDescTableExtended, true);
+            // When true, SSPs (adbc.databricks.ssp_*) are applied via post-open SET statements
+            // rather than CreateSession.session_confs — mirrors Thrift's behavior so callers
+            // who depend on the SET-statement path (e.g., for audit visibility or for SSPs
+            // the server validates differently in CreateSession vs SET) get the same semantics
+            // on SEA. JDBC has no equivalent flag — this is parity with the Thrift driver only.
+            _applySSPWithQueries = PropertyHelper.GetBooleanPropertyWithValidation(properties, DatabricksParameters.ApplySSPWithQueries, false);
 
             // Session configuration
             // Only supply catalog from connection properties when EnableMultipleCatalogSupport is true.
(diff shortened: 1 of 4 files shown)


csharp/src/StatementExecution/StatementExecutionConnection.cs
```diff
@@ -889,6 +901,48 @@ private static Dictionary<string, string> ExtractServerSideProperties(
...
+            if (serverSideProperties.Count == 0)
+            {
+                return;
+            }
+
+            foreach (var property in serverSideProperties)
+            {
+                // Backtick-escaped to match the Thrift path's SET escaping (DatabricksConnection.EscapeSqlString)
+                // — preserves values containing reserved characters while keeping a SET-compatible literal.
+                string escapedValue = "`" + property.Value.Replace("`", "``") + "`";
```
msrathore-db, line 929: Why do we need to replace this?

msrathore-db (approved):
LGTM

csharp/src/StatementExecution/StatementExecutionConnection.cs
eric-wang-1990: This is for escaping the values, e..g if the value contain whitespace or some restrained characters we need to escape.

PR #470, opened 2026-05-20
fix(csharp)!: consolidate SEA polling interval onto polltime_ms (PECO-3064) [SEA]

eric-wang-1990:
## What's Changed

The SEA path previously read only `adbc.databricks.rest.polling_interval_ms` and ignored `adbc.apache.statement.polltime_ms`. Rather than adding `polltime_ms` as a legacy alias (the original direction in cc4f69d), this PR **consolidates onto a single canonical key** — `adbc.apache.statement.polltime_ms` — for both protocols.

Resolution: SEA reads `adbc.apache.statement.polltime_ms` (default 1000 ms); Thrift continues to read the same key (default 100 ms). The property name is shared; only the protocol-specific default differs, since HTTP/JSON polls are heavier than Thrift RPCs.

### Breaking change

`adbc.databricks.rest.polling_interval_ms` is removed. Anyone setting that key on the SEA path must migrate to `adbc.apache.statement.polltime_ms`. SEA is opt-in (`adbc.databricks.protocol=rest`), so blast radius is bounded.

Title uses `fix(csharp)!:` per the conventional-commits guideline in CLAUDE.md.

## Why

PECO-3064 — M3 parameter gap, with a follow-up question raised in PR review: *"why have two variables for the same purpose?"* Consolidation eliminates the "two ways to do it" anti-pattern, matches JDBC's single-key convention (`asyncexecpollinterval`), and keeps protocol-specific defaults in code where they belong.

## Red → Green proof

E2E test `ExecuteQuery_PollTimeMs_DrivesSeaPollingCadence` sets `polltime_ms=2000`, `wait_timeout=0`, `enable_direct_results=false`, drains a `SELECT 1`, and asserts wall-clock ≥ 1800ms.

**Before fix** (against unmodified main):

```
Expected polltime_ms=2000 to drive SEA polling cadence, but query completed in 1364ms (expected >= 1800ms).
```

SEA fell back to the 1000 ms default; `polltime_ms` was silently dropped.

**After fix:** test passes (~2 s wall-clock); the full `StatementExecutionDriverE2ETests` class passes against the live `pecotesting` warehouse.

## Files touched

| File | Change |
|---|---|
| `csharp/src/DatabricksParameters.cs` | Delete `PollingInterval` constant |
| `csharp/src/StatementExecution/StatementExecutionConnection.cs` | Read only `ApacheParameters.PollTimeMilliseconds` (SEA default 1000 ms) |
| `csharp/test/Unit/DatabricksParametersTests.cs` | Remove `PollingInterval`-existence and REST-prefix assertions |
| `csharp/test/E2E/StatementExecution/StatementExecutionDriverE2ETests.cs` | Keep + rename the polltime_ms-drives-SEA-cadence test; delete the precedence and standalone-canonical tests (no precedence to test anymore) |
| `csharp/README.md` | Restructure Core Spark / Databricks-Specific / CloudFetch tables with `Thrift Default` / `SEA Default` columns. Auth, TLS, Proxy, Tracing tables get an "applies to both" header note. REST API Configuration noted as SEA-only, polling row removed |
| `csharp/doc/statement-execution-api-design.md` | Update polling-interval references to the consolidated key |

## Manual verification

- [x] E2E test `ExecuteQuery_PollTimeMs_DrivesSeaPollingCadence` passes against `pecotesting`
- [x] Full `StatementExecutionDriverE2ETests` class passes
- [x] `dotnet build` clean on all TFMs (0 warnings, 0 errors)
- [x] `DatabricksParametersTests` passes
- [ ] pre-commit (sandbox blocks hook env install; will run in CI)

PECO-3064

### DIFF
--- a/csharp/README.md
+++ b/csharp/README.md
@@ -286,7 +286,7 @@ The driver supports both Thrift/HiveServer2 protocol (default) and Statement Exe
 
 **REST API Configuration (SEA-only):**
 
-These properties apply only when `adbc.databricks.protocol=rest`. The cross-protocol polling interval is controlled by `adbc.apache.statement.polltime_ms` (see *Core Spark Properties* — SEA default `1000` ms).
+These properties apply only when `adbc.databricks.protocol=rest`.
 
 | Property | Description | SEA Default |
 |----------|-------------|-------------|
--- a/csharp/README.md
+++ b/csharp/README.md
@@ -191,23 +191,21 @@ The Databricks driver supports two protocols: **Thrift** (default, HiveServer2) 
 | Property | Description | Thrift Default | SEA Default |
 |----------|-------------|----------------|-------------|
 | `uri` | Full connection URI (alternative to host/port/path) | (required if no host/path) | (required if no host/path) |
-| `adbc.spark.type` | Server type: `http` | `http` | `http` |
+| `adbc.spark.type` | Server type: `http` | `http` | Not Supported (SEA is always HTTPS) |
 | `adbc.spark.host` | Hostname without scheme/port | (required) | (required) |
-| `adbc.spark.port` | Connection port | `443` | `443` |
+| `adbc.spark.port` | Connection port | `443` | Not Supported (HTTPS port from URI) |
 | `adbc.spark.path` | URI path on server | (required) | (required) |
-| `adbc.spark.token` | Token for token-based authentication | (none) | (none) |
-| `username` | Username for basic authentication | Not Supported | Not Supported |
-| `password` | Password for basic authentication | Not Supported | Not Supported |
+| `adbc.spark.token` | Token for token-based authentication | (none) | Not Supported (use `adbc.spark.auth_type=oauth` + `adbc.spark.access_token`) |
 | `adbc.spark.connect_timeout_ms` | Session establishment timeout (ms) | `30000` | Not Supported |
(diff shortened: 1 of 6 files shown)


csharp/README.md
```diff
@@ -276,15 +284,16 @@ The driver supports both Thrift/HiveServer2 protocol (default) and Statement Exe
 | `adbc.databricks.protocol` | Protocol: `thrift` or `rest` | `thrift` |
 | `adbc.databricks.warehouse_id` | Warehouse ID for query execution | (required for REST) |
 
-**REST API Configuration:**
+**REST API Configuration (SEA-only):**
 
-| Property | Description | Default |
-|----------|-------------|---------|
+These properties apply only when `adbc.databricks.protocol=rest`. The cross-protocol polling interval is controlled by `adbc.apache.statement.polltime_ms` (see *Core Spark Properties* — SEA default `1000` ms).
```
msrathore-db: Do we need to add this here?

msrathore-db (approved):
LGTM

csharp/README.md
eric-wang-1990: @copilot remove this part:  The cross-protocol polling interval is controlled by `adbc.apache.statement.polltime_ms` (see *Core Spark Properties* — SEA default `1000` ms).

PR #474, opened 2026-05-24
perf(csharp): reduce allocations and copies in result transfer

CurtHagenlocher:
## Summary

Four independent reductions in CPU/memory overhead on the result-transfer path. Each is small, localized, and covered by tests. Allocation figures below are from microbenchmarks (BenchmarkDotNet / `GC.GetAllocatedBytesForCurrentThread`).

| # | Area | Change | Effect |
|---|------|--------|--------|
| 1 | CloudFetch download (`CloudFetchDownloader`) | Read the HTTP body into one right-sized buffer instead of `MemoryStream` + `ToArray` | Removes a second full-size copy/alloc per chunk — ~2× faster, ½ allocation (8 MB: 16 → 8 MB/op) |
| 2 | SEA inline results (`StatementExecutionStatement`) | Feed chunks to `ArrowStreamReader` via a new forward-only `ConcatenatedReadOnlyMemoryStream` instead of `ToArray`-ing each chunk and concatenating into one buffer | Eliminates the whole-result allocation (8 MB/op → 0.055 KB/op) and the per-chunk copies |
| 3 | LZ4 decompression (`Lz4Utilities`) | Decompress into a pooled `RecyclableMemoryStream` + exact `ToArray` instead of a geometrically growing `MemoryStream` | Removes LOH growth churn and the over-sized result buffer (4 MB batch: 7.88 → 4.01 MB/op). Helps the inline Thrift reader **and** the SEA path |
| 4 | LZ4 work buffer (`CustomLZ4FrameReader`) | Return the decoder's work buffer without clearing it | ~83 µs/4 MB buffer saved; the decode overwrites it anyway |

Fixes #2 and #3 share `Lz4Utilities.DecompressLz4`; #3 threads the connection's existing `RecyclableMemoryStreamManager` to both the inline and SEA call sites.

## Why each is safe

- **2** — the SEA server splits a single Arrow IPC stream across N attachment chunks, so reading the segments back-to-back is byte-equivalent to the old concatenated buffer. `ConcatenatedReadOnlyMemoryStreamTest` verifies `ArrowStreamReader` returns identical batches/rows/values when the IPC bytes are split into 1 / 2 / 5 / 1000 segments (the 1000-segment case stresses cross-boundary reads).
- **3** — the returned array is an exact-size, GC-owned copy, so it stays valid for the `RecordBatch`'s lifetime after the pooled blocks are returned. Round-trip + equivalence asserted in `Lz4DecompressAllocationTest`.
- **4** — K4os bounds reads to the number of bytes it decoded, so stale buffer bytes are never surfaced. `Lz4DecompressNoStaleDataTest` proves this by decoding through a deliberately `0xFF`-poisoned, reused, never-cleared work buffer and still getting byte-exact output.

## Tests

New: `ConcatenatedReadOnlyMemoryStreamTest`, `Lz4DecompressNoStaleDataTest` (correctness), `CloudFetchBodyReadPerfTest`, `Lz4DecompressAllocationTest` (allocation characterization, loose ratio asserts). Updated `Lz4UtilitiesTests` for the new `DecompressLz4` signature. All pass locally on net8.0; the existing `CloudFetchDownloaderTest` / SEA `StatementExecution` suites stay green.

## Notes for reviewers

- The two allocation-characterization tests assert relative allocation (e.g. new ≤ old). If you'd prefer they not run in CI, I can gate them behind a trait or `SkippableFact`.
- **4 trade-off:** the previous `clearArray: true` could be read as memory-hygiene defense-in-depth. It was inconsistent anyway — the decompressed result also passes through `RecyclableMemoryStream` output blocks that are returned without clearing — so this is not a regression in end-to-end hygiene. If deliberate zeroing of sensitive data is wanted, that's a separate, end-to-end decision. Happy to drop #4 alone.
- The standalone `perf-analysis/` microbenchmark project used during analysis is intentionally **not** included in this PR.

### DIFF
--- a/csharp/src/CustomLZ4FrameReader.cs
+++ b/csharp/src/CustomLZ4FrameReader.cs
@@ -75,10 +75,14 @@ namespace AdbcDrivers.Databricks
         {
             if (buffer != null)
             {
-                // Clear the buffer to prevent stale data from previous decompressions
-                // from corrupting subsequent operations. The performance overhead (~1-2ms
-                // per 4MB buffer) is negligible compared to network I/O and decompression time.
-                _bufferPool.Return(buffer, clearArray: true);
+                // Return without clearing. This is the LZ4 decoder's internal block work buffer; K4os
+                // bounds every read to the number of bytes it actually decoded into the buffer, so stale
+                // bytes from a previous decompression are never surfaced to the output. This is verified
+                // by Lz4DecompressNoStaleDataTest, which decodes through a deliberately poisoned buffer
+                // and still gets byte-exact output. Upstream K4os likewise returns these buffers without
+                // clearing. Skipping the clear avoids zeroing a multi-MB buffer (~80us each, measured)
+                // that the next decode overwrites anyway.
+                _bufferPool.Return(buffer, clearArray: false);
             }
         }
     }
--- a/csharp/src/Lz4Utilities.cs
+++ b/csharp/src/Lz4Utilities.cs
@@ -47,50 +47,52 @@ namespace AdbcDrivers.Databricks
         /// Decompresses LZ4 compressed data into memory.
         /// </summary>
         /// <param name="compressedData">The compressed data bytes.</param>
+        /// <param name="memoryStreamManager">The RecyclableMemoryStreamManager used for the (pooled) decompression buffer.</param>
         /// <param name="bufferPool">The ArrayPool to use for buffer allocation (from DatabricksDatabase).</param>
         /// <returns>A ReadOnlyMemory containing the decompressed data.</returns>
         /// <exception cref="AdbcException">Thrown when decompression fails.</exception>
-        public static ReadOnlyMemory<byte> DecompressLz4(byte[] compressedData, ArrayPool<byte> bufferPool)
+        public static ReadOnlyMemory<byte> DecompressLz4(byte[] compressedData, RecyclableMemoryStreamManager memoryStreamManager, ArrayPool<byte> bufferPool)
         {
-            return DecompressLz4(compressedData, DefaultBufferSize, bufferPool);
+            return DecompressLz4(compressedData, DefaultBufferSize, memoryStreamManager, bufferPool);
         }
 
         /// <summary>
         /// Decompresses LZ4 compressed data into memory with a specified buffer size.
-        /// NOTE: This method uses regular MemoryStream (not RecyclableMemoryStream) because the buffer
-        /// must remain valid after the method returns for the caller to use.
-        /// Uses CustomLZ4DecoderStream with custom ArrayPool to efficiently pool 4MB+ buffers.
         /// </summary>
+        /// <remarks>
+        /// Decompresses into a pooled <see cref="RecyclableMemoryStream"/> rather than a plain
+        /// <see cref="MemoryStream"/>. A plain MemoryStream starts empty and doubles its backing array
+        /// as it grows, repeatedly reallocating and copying — landing on the Large Object Heap for the
+        /// multi-MB Arrow batches Databricks returns — and then leaves an over-sized buffer. The
+        /// recyclable stream instead writes into reused fixed-size blocks (no per-call LOH churn).
+        /// <see cref="RecyclableMemoryStream.ToArray"/> then produces a single exact-size, GC-owned
+        /// array, which is safe for the caller to retain after the pooled blocks are returned — the
+        /// Arrow <c>RecordBatch</c> built from it references this memory beyond this method's scope.
+        /// </remarks>
         /// <param name="compressedData">The compressed data bytes.</param>
         /// <param name="bufferSize">The buffer size to use for decompression operations.</param>
+        /// <param name="memoryStreamManager">The RecyclableMemoryStreamManager used for the (pooled) decompression buffer.</param>
         /// <param name="bufferPool">The ArrayPool to use for buffer allocation (from DatabricksDatabase).</param>
         /// <returns>A ReadOnlyMemory containing the decompressed data.</returns>
         /// <exception cref="AdbcException">Thrown when decompression fails.</exception>
-        public static ReadOnlyMemory<byte> DecompressLz4(byte[] compressedData, int bufferSize, ArrayPool<byte> bufferPool)
+        public static ReadOnlyMemory<byte> DecompressLz4(byte[] compressedData, int bufferSize, RecyclableMemoryStreamManager memoryStreamManager, ArrayPool<byte> bufferPool)
         {
             try
             {
-                // Use regular MemoryStream here because we return the buffer after disposal
-                // RecyclableMemoryStream would return the buffer to the pool, making it unsafe
-                using (var outputStream = new MemoryStream())
+                using (var outputStream = memoryStreamManager.GetStream())
+                using (var inputStream = new MemoryStream(compressedData))
+                using (var decompressor = new CustomLZ4DecoderStream(
+                    inputStream,
+                    descriptor => descriptor.CreateDecoder(),
+                    bufferPool,
+                    leaveOpen: false,
+                    interactive: false))
                 {
-                    using (var inputStream = new MemoryStream(compressedData))
-                    using (var decompressor = new CustomLZ4DecoderStream(
-                        inputStream,
-                        descriptor => descriptor.CreateDecoder(),
-                        bufferPool,
-                        leaveOpen: false,
-                        interactive: false))
-                    {
-                        decompressor.CopyTo(outputStream, bufferSize);
-                    }
+                    decompressor.CopyTo(outputStream, bufferSize);
 
-                    // Get the underlying buffer and its valid length without copying
-                    // The buffer remains valid after MemoryStream disposal since we hold a reference to it
-                    byte[] buffer = outputStream.GetBuffer();
-                    int length = (int)outputStream.Length;
-
-                    return new ReadOnlyMemory<byte>(buffer, 0, length);
+                    // Copy to an exact-size, GC-owned array before the pooled stream/blocks are
+                    // returned on dispose. The caller retains this beyond the stream's lifetime.
+                    return outputStream.ToArray();
                 }
             }
             catch (Exception ex)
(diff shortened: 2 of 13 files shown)


eric-wang-1990:
## Benchmark results: PR `41a6265` vs main `0f9f44c`

Ran the `CloudFetchRealE2EBenchmark` suite via `workflow_dispatch` on a base-repo branch (since fork PRs can't access secrets — see [run 26621093206](https://github.com/adbc-drivers/databricks/actions/runs/26621093206) for PR, [run 26621707849](https://github.com/adbc-drivers/databricks/actions/runs/26621707849) for main).

**Caveat**: the .NET 8.0 runs landed on different EPYC CPUs (`9V74` for PR, `7763` for main — GitHub-hosted runner pool). Times across them aren't directly comparable; allocation numbers are CPU-independent and trustworthy. The .NET Framework 4.7.2 runs both hit `EPYC 7763` Windows, so those times are apples-to-apples.

### .NET Framework 4.7.2 (clean — same CPU)

| query | Mean (PR → main) | Δ time | Allocated (PR → main) | Δ alloc |
|---|---|---:|---|---:|
| customer | 743 ms → 1414 ms | **−47.5%** 🟢 | 33.4 MB → 38.9 MB | −14.2% 🟢 |
| catalog_sales | 4.22 s → 5.36 s | **−21.2%** 🟢 | 424 MB → 546 MB | **−22.4%** 🟢 |
| web_sales | 3.24 s → 3.82 s | −14.9% 🟢 | 347 MB → 407 MB | −14.9% 🟢 |
| wide_sales_analysis | 13.84 s → 15.52 s | −10.8% 🟢 | 2383 MB → 2928 MB | −18.6% 🟢 |
| store_sales_numeric | 12.50 s → 13.22 s | −5.4% 🟢 | 255 MB → 405 MB | **−37.2%** 🟢 |
| inventory | 45.14 s → 45.73 s | −1.3% 🟢 | 187 MB → 264 MB | **−28.9%** 🟢 |
| sales(...)tamps | 13.23 s → 13.37 s | −1.0% 🟢 | 119 MB → 150 MB | −20.7% 🟢 |

PR wins on **every query** in both axes. Allocation reductions 14–37%; time wins range from negligible to 47%.

### .NET 8.0 (allocation only — times skewed by CPU mismatch)

| query | Allocated (PR → main) | Δ alloc |
|---|---|---:|
| wide_sales_analysis | 612 MB → 1170 MB | **−47.7%** 🟢 |
| catalog_sales | 158 MB → 286 MB | **−44.7%** 🟢 |
| store_sales_numeric | 194 MB → 343 MB | **−43.4%** 🟢 |
| inventory | 162 MB → 242 MB | −33.1% 🟢 |
| customer | 27.9 MB → 41.2 MB | −32.2% 🟢 |
| web_sales | 103.6 MB → 150.6 MB | −31.2% 🟢 |
| sales(...)tamps | 80.8 MB → 113.2 MB | −28.6% 🟢 |

Across all 7 queries, allocation drops **28–48%**, consistent with the PR's claims (CloudFetch chunk 16→8 MB; SEA inline 8 MB→0.055 KB; LZ4 7.88→4.01 MB per batch).

### Conclusion

The PR delivers what it claims:
- **.NET 4.7.2**: faster *and* lighter on every query — strongest, cleanest evidence.
- **.NET 8.0**: allocation savings are real and large (28–48%). Time comparison is unreliable due to runner CPU mismatch — would need both runs pinned to the same runner generation for a fair time delta.

One number worth flagging for follow-up: **`wide_sales_analysis` on .NET 4.7.2** still allocates **2.38 GB** post-PR for ~2.7M rows. That's ~5× the next-heaviest query. Not blocking, but worth a future profile.

Full artifacts: [run 26621093206](https://github.com/adbc-drivers/databricks/actions/runs/26621093206) (PR), [run 26621707849](https://github.com/adbc-drivers/databricks/actions/runs/26621707849) (main).


---

eric-wang-1990 (approved):

csharp/src/CustomLZ4FrameReader.cs
```diff
@@ -75,10 +75,14 @@ protected override void ReleaseBuffer(byte[] buffer)
...
-                // per 4MB buffer) is negligible compared to network I/O and decompression time.
-                _bufferPool.Return(buffer, clearArray: true);
+                // Return without clearing. This is the LZ4 decoder's internal block work buffer; K4os
+                // bounds every read to the number of bytes it actually decoded into the buffer, so stale
+                // bytes from a previous decompression are never surfaced to the output. This is verified
+                // by Lz4DecompressNoStaleDataTest, which decodes through a deliberately poisoned buffer
+                // and still gets byte-exact output. Upstream K4os likewise returns these buffers without
+                // clearing. Skipping the clear avoids zeroing a multi-MB buffer (~80us each, measured)
+                // that the next decode overwrites anyway.
+                _bufferPool.Return(buffer, clearArray: false);
```
eric-wang-1990, line 85: Worth a note on the containment scope: `_bufferPool` is `connection.Lz4BufferPool`, but per `DatabricksDatabase.cs:56-62` that pool is **database-scoped, not connection-scoped** — every connection from one `DatabricksDatabase` shares the same `ArrayPool<byte>` instance. So if a host application opens connections from one Database on behalf of multiple end users (multi-tenant service), work buffers can be rented by a different user's decode after this `Return`.

K4os's contract still protects normal-path correctness (LZ4 match references are bounded to within the current frame's already-decoded output; the work buffer is internal scratch and decompressed bytes go to a separate destination), so the empirical `Lz4DecompressNoStaleDataTest` result is consistent with that. The risk is purely defense-in-depth: `clearArray: true` previously hedged against any future K4os bug that over-read into stale bytes; that hedge is now gone, AND the pool spans tenants in a multi-tenant host.

Not blocking. Two options for a small follow-up:
1. Document this constraint near `Lz4BufferPool`'s declaration: "must remain database-scoped — do not promote to a static/global pool — because work buffers are returned without clearing." Belt-and-suspenders for future maintainers.
2. Or restore `clearArray: true`. PR description quotes ~83µs/4MB — cheap relative to the body-read win, and end-to-end query latency is dominated by I/O + decompression rather than buffer zeroing.

For single-user host applications (PowerBI, Tableau, Excel) this is irrelevant. Flagging because the comment above implies pool isolation that doesn't actually exist at the Database level.

---

csharp/src/StatementExecution/ConcatenatedReadOnlyMemoryStream.cs
```diff
@@ -0,0 +1,137 @@
...
+        public override bool CanWrite => false;
+        public override long Length => _length;
+
+        public override long Position
+        {
+            get => _position;
+            set => throw new NotSupportedException("ConcatenatedReadOnlyMemoryStream is forward-only.");
+        }
+
+        public override int Read(byte[] buffer, int offset, int count)
```
eric-wang-1990, line 71: Minor / defense-in-depth: no `_disposed` flag, so `Read`/`ReadAsync` after `Dispose()` will silently keep walking `_segments` instead of throwing `ObjectDisposedException`. In practice the only caller is `InlineArrowStreamReader.Dispose`, which disposes `_streamReader` before `_dataStream`, so the ordering is fine today and the class is `internal sealed` so the surface is bounded.

Not blocking. If you want belt-and-suspenders against a future refactor: a one-line `_disposed` guard at the top of `ReadCore` plus an override of `Dispose(bool)` that sets it. Otherwise feel free to ignore.

---

csharp/src/Lz4Utilities.cs
```diff
@@ -47,50 +47,52 @@ internal static class Lz4Utilities
...
-                    // Get the underlying buffer and its valid length without copying
-                    // The buffer remains valid after MemoryStream disposal since we hold a reference to it
-                    byte[] buffer = outputStream.GetBuffer();
-                    int length = (int)outputStream.Length;
+                    decompressor.CopyTo(outputStream, bufferSize);
 
-                    return new ReadOnlyMemory<byte>(buffer, 0, length);
+                    // Copy to an exact-size, GC-owned array before the pooled stream/blocks are
+                    // returned on dispose. The caller retains this beyond the stream's lifetime.
+                    return outputStream.ToArray();
```
eric-wang-1990, line 95: FYI / non-blocking: the backing array of the returned `ReadOnlyMemory<byte>` changes shape with this PR. Old code returned `new ReadOnlyMemory<byte>(buffer, 0, length)` where `buffer` was the `MemoryStream`'s oversized internal buffer (so `seg.Array.Length ≥ seg.Count`). After this change, `RecyclableMemoryStream.ToArray()` returns a new array sized exactly to the decompressed length (`seg.Array.Length == seg.Count`).

Any caller that uses `MemoryMarshal.TryGetArray(rom, out var seg)` and then inspects `seg.Array.Length` (rather than `seg.Count`) would observe the tighter size. I grepped the codebase and didn't find any such caller — Arrow's deserializer respects the `.Length` view — so this should be invisible in practice. Just calling it out in case future code assumes the looser sizing.

---

csharp/src/CustomLZ4FrameReader.cs
eric-wang-1990: Addressed in b59072e (option 1 — documented the constraint, kept the perf win). Added a `<remarks>` block on the `Lz4BufferPool` declaration in `DatabricksDatabase.cs` stating the pool MUST remain database-scoped and must not be promoted to a static/global pool, because `CustomLZ4FrameReader.ReleaseBuffer` returns work buffers uncleared — so the pool's reuse boundary is also the cross-tenant data-isolation boundary. Belt-and-suspenders for future maintainers without giving back the ~83µs/4MB zeroing cost.

---

csharp/src/Lz4Utilities.cs
eric-wang-1990: No change needed — this was an FYI. Confirmed no caller inspects `seg.Array.Length` (vs `seg.Count`), so the tighter `ToArray()` sizing is invisible in practice. Resolving.

---

csharp/src/StatementExecution/ConcatenatedReadOnlyMemoryStream.cs
eric-wang-1990: Decided to skip the `_disposed` guard. The class is `internal sealed`, the only caller (`InlineArrowStreamReader.Dispose`) disposes `_streamReader` before `_dataStream`, and the suggestion was explicitly defense-in-depth against a hypothetical future refactor. Not worth the extra surface today. Resolving.

---

PR #490, opened 2026-05-28
tracing(csharp): tag Execute spans with query_timeout_s

eric-wang-1990:
## Motivation

When a query times out (e.g. `QueryTimeoutSeconds=5` against a long
count-over-join), the trace correctly records `Status=Error` on
`HiveServer2Statement.ExecuteQueryAsyncInternal` /
`ExecuteStatementAsync` with `exception.type=TTransportException` and
the cancel-operation event chain. But the configured timeout value
(`adbc.apache.statement.query_timeout_s`) only lived on the
`CreateSessionRequest` of a *different* trace. The timing-out span
carried no threshold tag, so an operator reading just the failing span
had to do duration math ("Duration=5.30s -> probably a 5s timeout") to
recover the configured threshold. See issue #478.

## What changed

- `csharp/src/DatabricksStatement.cs` — in the
  `SetStatementProperties(TExecuteStatementReq)` override (which runs
  inside the `HiveServer2Statement.ExecuteStatementAsync` activity),
  emit `Activity.Current?.SetTag(ApacheParameters.QueryTimeoutSeconds,
  QueryTimeoutSeconds.ToString(InvariantCulture))`. Matches the
  existing tag-naming convention exactly
  (`adbc.apache.statement.query_timeout_s`) so debugger/log queries
  keyed on this name work uniformly across CreateSession and Execute
  spans.

## Test added

`StatementTimeout_TagsExecuteSpanWithQueryTimeoutSeconds_Issue478` in
`csharp/test/E2E/StatementTests.cs`. Wires an `ActivityListener`
against the Databricks/HiveServer2 activity sources, configures
`QueryTimeoutSeconds=5`, runs the existing
`LongRunningStatementTimeoutTestData.LongRunningQuery`, asserts
`Assert.Throws<TimeoutException>(...)`, and then asserts that at
least one captured Execute span carries
`adbc.apache.statement.query_timeout_s=5`. Before the fix this test
fails with the expected diagnostic ("Inspected 2 Execute span(s);
their tags: [...] -- no `query_timeout_s` tag"); after the fix it
passes.

The test was committed first as RED (failing), then the production
fix was applied to turn it GREEN — see commits on this branch.

## Regression check

`StatementTests` class: 95 passed / 3 skipped / 0 failed against
`sf10-pat`.

Closes #478

This pull request and its description were written by Isaac.

Issue #478, opened 2026-05-27
tracing(csharp): query timeout threshold not on the failing Execute span


eric-wang-1990:
**Severity:** Medium

## Observation
When `StatementTests.StatementTimeoutTest` triggers a `TimeoutException` (e.g. `QueryTimeoutSeconds=5` against a long count-over-join), the trace correctly captures `Status=Error` on `HiveServer2Statement.ExecuteQueryAsyncInternal` with `exception.type=TTransportException`, `exception.message="The operation was canceled."`, and event chain `cancel_operation.starting` → `cancel_operation.completed` (with `db.response.status_code=SUCCESS_STATUS` — server acked the cancel).

**But** the configured timeout value (`adbc.apache.statement.query_timeout_s = "5"`) lives only on the `CreateSessionRequest` of a *different* TraceId. The timing-out span carries no threshold tag. An operator reading just the failing span sees "Duration=5.30s" and has to do math to infer "this was a 5s timeout."

## Suggested fix
Copy the live `query_timeout_s` (and other relevant statement-level config like `polltime_ms`, `connect_timeout_ms`) onto `HiveServer2Statement.ExecuteStatementAsync` and `ExecuteQueryAsyncInternal` spans, not just the session-creation span.

## Evidence
- `Traces.StatementTests.20260527_144246/AdbcDrivers.Databricks-trace-2026-05-27-21-45-30-448111-28640.log:10` — error span with no `query_timeout_s` tag
- Same file, line 1 — `CreateSessionRequest` of a separate trace carries `adbc.apache.statement.query_timeout_s = "5"`
- Across all 93 error spans in StatementTests archive: zero carry the timeout tag

PR #491, opened 2026-05-28
tracing(csharp): omit db.response.returned_rows tag when -1 sentinel

### LINKED ISSUE #484: tracing(csharp): db.response.returned_rows = -1 on UPDATE spans is misleading
**Severity:** Low

## Observation
`HiveServer2Statement.ExecuteUpdateAsyncInternal` always reports `db.response.returned_rows = -1` on every UPDATE-flavored statement (CREATE, INSERT, UPDATE, DELETE). Verified across all archives — never another value. This appears to be a sentinel because the Thrift server doesn't return affected-row count for DML statements via this pathway.

`-1` reads as "we measured zero rows" or "we don't know" depending on the consumer, and in many APMs it sorts alongside legitimate row counts.

## Suggested fix
Either omit the tag entirely on UPDATE paths, OR set it explicitly to `null`, OR rename the tag to `db.response.affected_rows` and emit a dedicated value (-1 → meaningful "unknown" semantics in API docs).

## Evidence
- `Traces.StatementTests.20260527_144246/` — all `ExecuteUpdateAsyncInternal` spans carry `db.response.returned_rows = -1`
- `Traces.ClientTests.20260527_144710/` — same

eric-wang-1990:
## Motivation

`HiveServer2Statement.ExecuteUpdateAsyncInternal` (in the
hiveserver2 submodule) always tags UPDATE-flavored spans
(CREATE/INSERT/UPDATE/DELETE/DROP) with
`db.response.returned_rows = -1` when the Thrift response carries no
`num_affected_rows` column — i.e., for DDL paths where the server
genuinely doesn't return an affected-row count. The `-1` is a sentinel
masquerading as a row count: in APM dashboards it sorts alongside
legitimate counts, and consumers naturally misread it as "we measured
zero" or "we don't know". See issue #484.

## What changed

- `csharp/src/DatabricksStatement.cs` — adds a one-time process-global
  `ActivityListener` registered in an eager static initializer. The
  listener inspects every `*.ExecuteUpdateAsyncInternal` span on
  `ActivityStopped` and, when `db.response.returned_rows` carries the
  `-1` sentinel, removes the tag entirely via `SetTag(key, null)`. Tag
  removal during `ActivityStopped` is safe: `Activity.IsStopped` flips
  to `true` only after all `ActivityStopped` callbacks return.
- The listener's `Sample` callback returns `PropagationData` (the
  minimum sampling level that still delivers `ActivityStopped`
  notifications). Effective recording level across listeners is the
  max, so this does NOT degrade other listeners that ask for
  `AllDataAndRecorded`.
- An explicit `static DatabricksStatement()` constructor forces eager
  static-field initialization (defeating `beforeFieldInit` lazy
  semantics), guaranteeing the listener is registered before any
  per-connection `FileActivityListener` subscribes — so file exporters
  also never see the `-1` value.
- When the Thrift response does carry a real `num_affected_rows` (>= 0),
  the existing submodule code path tags the span with that value
  unchanged.

The submodule is read-only from this repo (same constraint as
PR #490 / issue #478), so the listener-based scrub is the least
invasive in-driver fix.

## Test added

`ExecuteUpdate_DoesNotEmitMinusOneReturnedRows_Issue484` in
`csharp/test/E2E/StatementTests.cs`. Wires an `ActivityListener`
over the Databricks/HiveServer2 activity sources, runs
CREATE/INSERT/UPDATE/DROP through `ExecuteUpdateAsyncInternal` against
a transient table, then asserts that no captured Execute span carries
`db.response.returned_rows = -1`. Without the fix the test fails with
two offending CREATE/DROP spans carrying the sentinel (INSERT/UPDATE
have positive affected_rows so they're naturally unaffected). With the
fix the test passes.

The test was committed first as RED (failing), then the production
fix turned it GREEN — see commits on this branch.

## Regression check

- `StatementTests` class: 95 passed / 3 skipped / 0 failed against
  the configured Thrift workspace.
- `DriverTests` class: 35 passed / 0 failed.

Closes #484

### DIFF
--- a/csharp/src/DatabricksStatement.cs
+++ b/csharp/src/DatabricksStatement.cs
@@ -24,6 +24,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


eric-wang-1990:
Superseded — the proper fix belongs in the hiveserver2 submodule at HiveServer2Statement.cs:242 (3-line guard inside the existing TraceActivityAsync lambda). Process-global ActivityListener is an unnecessary workaround now that we have write access on the submodule. A new submodule PR will replace this; parent repo will then bump the submodule pointer + keep the regression test (without the listener).

PR #492, opened 2026-05-28
fix(csharp): bracket close_operation event for measurable latency

### LINKED ISSUE #489: tracing(csharp): composite_reader.close_operation event shares timestamp with disposing — close latency unmeasurable
**Severity:** Low

## Observation
`DatabricksCompositeReader.Dispose` emits an event sequence:

```
composite_reader.disposing            t=21:31:29.768505
composite_reader.close_operation      t=21:31:29.768505   ← same timestamp!
composite_reader.disposing_active_reader t=21:31:29.768510
composite_reader.disposed             t=21:31:29.768510
```

`composite_reader.disposing` and `composite_reader.close_operation` are emitted back-to-back with identical timestamps. So the *latency* of the close-operation RPC against the Thrift server (which is what `close_operation` is meant to mark) is NOT separately measurable from the rest of the Dispose sequence. The parent Dispose span carries a total Duration, but that includes everything.

If `close_operation` is genuinely fast (sub-µs) this is fine. If it's slow (network round-trip), the trace can't tell you — which defeats the purpose of having a dedicated event.

## Suggested fix
Either:
- Emit `composite_reader.close_operation.started` and `composite_reader.close_operation.completed` as separate events bracketing the actual RPC call, OR
- Replace the event with a child span `CompositeReader.CloseOperation` so it has a real Duration.

## Evidence
- `Traces.CloseOperationE2ETest.20260527_144932/` — all 3 disposes show same-timestamp events
- `Traces.DriverTests.20260527_153928/` — same pattern across all reader disposes

eric-wang-1990:
## Motivation

`DatabricksCompositeReader.Dispose` previously emitted a single
`composite_reader.close_operation` event immediately before invoking
the Thrift `TCloseOperationReq` RPC. Because the event fired before
the call, its timestamp was effectively identical to the surrounding
`composite_reader.disposing` event (sample observed in issue #489:
both at `t=21:31:29.768505`). The RPC's wall-clock cost was therefore
invisible in the trace — defeating the whole purpose of having a
dedicated event for the close call. A repro on `sf10/pecotesting`
showed the next event (`composite_reader.disposing_active_reader`)
firing ~142ms later, so the latency is non-trivial and worth
capturing.

## What changed

- `csharp/src/Reader/DatabricksCompositeReader.cs` — in
  `Dispose(bool)`, replace the single
  `composite_reader.close_operation` event with a pair:
  - `composite_reader.close_operation.started` BEFORE the RPC
  - `composite_reader.close_operation.completed` AFTER the RPC
    (emitted in the `finally` block so it always fires; tagged
    `error=true/false` so consumers can distinguish a clean close
    from a thrown one).

  This mirrors the existing event-pair convention used in
  `HiveServer2Statement.CancelOperationAsync`
  (`db.operation.cancel_operation.starting` / `.completed`). The
  delta between the two timestamps now captures the RPC's latency
  directly in the trace.

- `csharp/test/E2E/CloseOperationE2ETest.cs` — the existing
  `DisposeEmitsCloseOperationEvent` theory was asserting on the bare
  `composite_reader.close_operation` name; updated it to assert on
  the new `.started` event name since the bare event no longer
  exists.

## Test added

`DisposeCloseOperation_EmitsStartedAndCompletedWithDistinctTimestamps_Issue489`
in `csharp/test/E2E/CloseOperationE2ETest.cs`. Wires the same
`ActivityListener` the existing test uses but additionally captures
full `ActivityEvent` objects (with timestamps), runs a query that
triggers a non-inline close RPC, disposes the reader and statement,
and asserts:

1. `composite_reader.close_operation.started` is present
2. `composite_reader.close_operation.completed` is present
3. `completed.Timestamp > started.Timestamp` (strict — any positive
   delta means the RPC is now measurable)

The test was committed RED first; against the unpatched driver it
fails with the diagnostic `Got events: [composite_reader.disposing,
composite_reader.close_operation, composite_reader.disposing_active_reader,
composite_reader.disposed]` (sample timestamps:
`disposing @ 06:54:05.501959`, `close_operation @ 06:54:05.501960` —
1µs apart, then `disposing_active_reader @ 06:54:05.643804` — 142ms
later, i.e. the RPC took ~142ms but the old single-event design
showed 0). After the fix the pair brackets the RPC cleanly and the
test passes.

## Regression check

- `CloseOperationE2ETest` (4 tests, including the new one): all pass.
- `DriverTests`: 35/35 pass.
- `StatementTests`: 94 passed / 3 skipped / 0 failed.

All against `sf10/pecotesting`, Thrift protocol.

Closes #489

### DIFF
--- a/csharp/src/Reader/DatabricksCompositeReader.cs
+++ b/csharp/src/Reader/DatabricksCompositeReader.cs
@@ -252,7 +252,18 @@ namespace AdbcDrivers.Databricks.Reader
                             // Always close the operation here at the composite level.
                             // CloudFetchReader is protocol-agnostic and does not send CloseOperation,
                             // so we must not rely on the contained reader to do it.
-                            activity?.AddEvent("composite_reader.close_operation");
+                            //
+                            // Issue #489: bracket the Thrift TCloseOperationReq RPC with a pair
+                            // of events ("...started" before, "...completed" after) so the
+                            // timestamp delta between them captures the RPC's wall-clock
+                            // latency in the trace. Previously a single
+                            // "composite_reader.close_operation" event was emitted before the
+                            // call, which made its timestamp identical to the surrounding
+                            // "composite_reader.disposing" event and hid the RPC's cost.
+                            // Mirrors the cancel_operation event-pair pattern used in
+                            // HiveServer2Statement.CancelOperationAsync ("...starting" /
+                            // "...completed").
+                            activity?.AddEvent("composite_reader.close_operation.started");
                             // Time the TCloseOperationReq RPC so CLOSE_STATEMENT telemetry
                             // emitted later from DatabricksStatement.Dispose can report the
                             // actual server-side close latency (PECO-2991).
@@ -276,6 +287,13 @@ namespace AdbcDrivers.Databricks.Reader
                                     dbxStatement.CloseStatementRpcLatencyMs = closeStopwatch.ElapsedMilliseconds;
                                     dbxStatement.CloseStatementRpcError = closeError;
                                 }
+                                // Emit the matching "completed" event regardless of whether
+                                // the RPC threw so the trace always records when the close
+                                // call returned. Tag with error=true on failure so consumers
+                                // can distinguish a clean close from a thrown one.
+                                activity?.AddEvent(
+                                    "composite_reader.close_operation.completed",
+                                    [new("error", closeError != null)]);
                             }
                             if (_activeReader != null)
                             {
--- a/csharp/src/Reader/DatabricksCompositeReader.cs
+++ b/csharp/src/Reader/DatabricksCompositeReader.cs
@@ -290,10 +290,12 @@ namespace AdbcDrivers.Databricks.Reader
                                 }
                                 // Emit the matching "completed" event regardless of whether
                                 // the RPC threw so the trace always records when the close
-                                // call returned. Error details, if any, are surfaced via
-                                // AddException above (matches the convention used in
-                                // HiveServer2Statement.CancelOperationAsync).
-                                activity?.AddEvent("composite_reader.close_operation.completed");
+                                // call returned. Error details are also surfaced via
+                                // AddException above; the tag here lets consumers filter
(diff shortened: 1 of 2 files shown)


csharp/src/Reader/DatabricksCompositeReader.cs
```diff
@@ -276,6 +287,13 @@ protected override void Dispose(bool disposing)
                                     dbxStatement.CloseStatementRpcLatencyMs = closeStopwatch.ElapsedMilliseconds;
                                     dbxStatement.CloseStatementRpcError = closeError;
                                 }
+                                // Emit the matching "completed" event regardless of whether
+                                // the RPC threw so the trace always records when the close
+                                // call returned. Tag with error=true on failure so consumers
+                                // can distinguish a clean close from a thrown one.
+                                activity?.AddEvent(
+                                    "composite_reader.close_operation.completed",
+                                    [new("error", closeError != null)]);
```
eric-wang-1990: This probably does not work in Net 472.
Do we have AddError? Or can we just put error string there if any? Look for other activity example to surface errors

eric-wang-1990: Good catch — switched to `activity?.AddException(ex)` in the catch block (bebd49e), matching the convention in `HiveServer2Statement.CancelOperationAsync`, `DatabricksReader`, and `DatabricksConnection`. Dropped the bool `error` tag entirely. The `.completed` event still fires in `finally` so the latency measurement is preserved; error details are now surfaced via `AddException` which works reliably across net472/netstandard2.0/net8.0.

---

msrathore-db (approved):
Verified live: all 4 `CloseOperationE2ETest` cases pass, including the new distinct-timestamps test. The `.started`/`.completed` pairing in the `finally` block correctly brackets the RPC. LGTM.

Minor (non-blocking): the description says `.completed` is "tagged `error=true/false`", but the code emits it with no tags — errors are recorded via `activity?.AddException(ex)` instead. Worth aligning the description (or adding the tag) so consumers don't look for an `error` tag on the event that isn't there.

PR #495, opened 2026-05-28
fix(csharp): tag download_slot_acquired with wait_duration_ms

### LINKED ISSUE #483: tracing(csharp): cloudfetch.download_slot_acquired carries no wait_duration — slot contention invisible
**Severity:** Low

## Observation
`DownloadFilesAsync` emits a `cloudfetch.download_slot_acquired` event every time a download slot is acquired from the parallelism pool. Under `CloudFetchStressTests` (24-file / 241MB downloads) the event fires 252 times.

But the event carries **no `wait_duration_ms` tag** — so it's impossible to tell whether a slot was acquired immediately or whether the call blocked waiting for another download to finish. Under high parallelism, this is exactly the metric you'd want to diagnose "downloads are slow because they're queued."

## Suggested fix
Capture `Stopwatch` around `slot.WaitAsync()` and add `wait_duration_ms` tag to the `cloudfetch.download_slot_acquired` event. Also consider `queue_depth` (count of pending downloads).

## Evidence
- `Traces.CloudFetchStressTests.20260527_154205/` — 252 `cloudfetch.download_slot_acquired` events, none with wait_duration
- Tier 1 finding: CloudFetchMemoryBufferManager has no dedicated span / event for back-pressure either

eric-wang-1990:
## Summary
- Closes #483
- `cloudfetch.download_slot_acquired` previously fired with only `chunk_index` and `is_sequential_mode`, so slot contention was invisible in traces. Under `CloudFetchStressTests` the event fired 252 times with no way to distinguish "acquired instantly" from "queued behind another download."
- This PR wraps the `SemaphoreSlim.WaitAsync` in a `Stopwatch` and attaches `wait_duration_ms` (long, milliseconds) plus `queue_depth` (count of pending downloads at acquisition) to the event.

## Why both tags?
The issue calls for `wait_duration_ms` as the primary signal and lists `queue_depth` as an optional follow-up. I included both because `_downloadQueue.Count` was already used in the surrounding code (see `cloudfetch.download_loop_start` at line 326) — capturing it under the stopwatch snapshot is a single extra field load with no new locking, so leaving it out would be more work to justify than to keep.

## Test (RED -> GREEN)
- New test `CloudFetchE2ETest.DownloadSlotAcquired_EmitsWaitDurationMs_Issue483` runs a 24-file CloudFetch download (`SELECT * FROM main.tpcds_sf100_delta.store_sales LIMIT 1000000` with `MaxBytesPerFile=10MB`) and asserts that every captured `cloudfetch.download_slot_acquired` event carries a non-negative integer `wait_duration_ms` tag.
- RED: pre-fix run captures 24 events, all with only `chunk_index`/`is_sequential_mode`, no `wait_duration_ms`.
- GREEN: post-fix run passes. Actual values from the verified run show real backpressure (chunks 0-2 acquire immediately at 0ms, then later chunks wait 50-605ms behind the default `_maxParallelDownloads=3` cap):
  ```
  wait_duration_ms = 0    queue_depth = 0
  wait_duration_ms = 0    queue_depth = 4
  wait_duration_ms = 605  queue_depth = 3
  wait_duration_ms = 94   queue_depth = 3
  ...
  ```
- Full `CloudFetchE2ETest` suite (21 tests) passes after the fix.

## Files changed
- `csharp/src/Reader/CloudFetch/CloudFetchDownloader.cs` (+23/-2)
- `csharp/test/E2E/CloudFetchE2ETest.cs` (+139)


msrathore-db (approved):
Verified live: test passes; `wait_duration_ms` and `queue_depth` are emitted on the `download_slot_acquired` event. Clean change. LGTM. (Tiny nit, non-blocking: `queue_depth` is snapshotted just before `WaitAsync` rather than at acquisition — cosmetic.)

PR #497, opened 2026-05-28
fix(csharp): emit retry telemetry from RetryHttpHandler

### LINKED ISSUE #479: tracing(csharp): RetryHttpHandler emits no telemetry — retries are invisible
**Severity:** Medium

## Observation
The `RetryHttpHandler` (`csharp/src/Auth/RetryHttpHandler.cs`) runs on every HTTP send — confirmed by stack-trace evidence in `Traces.DriverResilienceTests.20260527_152945/`:

```
at AdbcDrivers.Databricks.RetryHttpHandler.<>c__DisplayClass11_0.<<SendAsync>b__0>d.MoveNext()
   in .../RetryHttpHandler.cs:line 135
```

But the handler emits **zero** telemetry. No `retry.attempt`, `retry.count`, `retry.delay_ms`, `retry.reason`, `retry.exhausted` tags or events anywhere. Verified by:

1. `grep -ric "retry" *.log` across all 19 trace files in `DriverResilienceTests` + the dedicated `Traces.RetryCountTests.20260527_152556/` archive = **0 hits** (other than the config tag `adbc.spark.temporarily_unavailable_retry`).
2. Looking for the "retry shape" pattern of multiple sibling `SendAsync` spans under one parent — `grep "SendAsync"` shows **0 parents have more than 1 SendAsync child**, even on error paths.

A customer asking "is my failure flaky (retried then succeeded) or hard (no retry available)?" cannot answer from the trace alone.

## Caveat
`hive2.thrift.poll_attempts` event-tag DOES exist on `HiveServer2Connection.PollForResponseAsync` — but that's Thrift operation-status polling, NOT HTTP-level retry.

## Suggested fix
In `RetryHttpHandler.SendAsync`, around the retry loop, call `Activity.Current?.AddEvent("retry.attempt", attempt_n, reason, delay_ms)` and set `retry.total_attempts` / `retry.exhausted` tags on the wrapping `SendAsync` span on completion.

## Evidence
- `Traces.DriverResilienceTests.20260527_152945/` — exception stacktrace traverses RetryHttpHandler.cs:135 but no retry telemetry
- `Traces.RetryCountTests.20260527_152556/` — dedicated test for retry counts, contains no `retry.*` tag or event anywhere

eric-wang-1990:
## Summary

Issue #479: `RetryHttpHandler` runs on every HTTP send and retries on Retry-After (503/429) responses and transient transport errors, but historically emitted zero per-attempt telemetry. A user reading their own driver logs asking "did my failure retry then succeed, or did it fail without any retry?" couldn't answer from the trace alone.

This PR makes each retry visible as an event on the `SendAsync` activity.

### Event vocabulary

On the activity that wraps each `RetryHttpHandler.SendAsync` call — one `retry.attempt` event per *retry* (NOT emitted on first-try success):

- `attempt_number` (int) — 1-indexed retry counter, sourced directly from the existing `attemptCount`
- `delay_ms` (long) — milliseconds the handler will sleep before re-sending
- `reason` (string) — one of:
  - `retry_after_<status>` (e.g. `retry_after_503`, `retry_after_429`)
  - `retry_after_<status>_invalid_header` — header present but unparseable
  - `retry_after_<status>_no_header` — retryable status, no Retry-After
  - `transport_error_<ExceptionTypeName>` — transient transport failure

### What's intentionally NOT included

No `retry.total_attempts` / `retry.exhausted` summary tags. The target consumer is a user reading their own driver logs, so dashboard-shaped aggregation tags aren't worth the code complexity — the per-retry events already carry the diagnostic surface ("how many times did this retry, and why"). The pre-existing `http.retry.total_attempts` tag on the exhaustion-break paths (already on `main` before this PR) is untouched.

### Files changed

- `csharp/src/RetryHttpHandler.cs` — emits `retry.attempt` events inside the existing `TraceActivityAsync` block, both on the transport-error retry path and the Retry-After 503/429 retry path. Uses the existing `attemptCount` for the event's `attempt_number` (no new counters).
- `csharp/test/Unit/RetryHttpHandlerTest.cs` — two new tests assert on the events emitted from the `SendAsync` activity:
  - `RetryTelemetry_NoRetry_EmitsNoEvents_Issue479` — happy path: no events on first-try success.
  - `RetryTelemetry_RetryAfter_EmitsAttemptEvents_Issue479` — `[Theory]` over 503 and 429: every retry emits an event with monotonically increasing `attempt_number`, `delay_ms >= 1000` (Retry-After: 1), and a `reason` containing the status code so 429 throttles are distinguishable from 503s.

## Test plan

- [x] `dotnet test --filter "FullyQualifiedName~RetryTelemetry"` — 3/3 pass (1 `[Fact]` + 2 `[Theory]` cases)
- [x] `dotnet test --filter "FullyQualifiedName~RetryHttpHandlerTest"` — full retry suite passes
- [x] Build clean across netstandard2.0, net472, net8.0

Closes #479

### DIFF
--- a/csharp/src/RetryHttpHandler.cs
+++ b/csharp/src/RetryHttpHandler.cs
@@ -109,243 +109,202 @@ namespace AdbcDrivers.Databricks
             int totalTooManyRequestsRetrySeconds = 0;
             int totalTransportErrorRetrySeconds = 0;
 
-            // Issue #479: telemetry counters that survive across the retry
-            // loop so the summary tags (retry.total_attempts / retry.exhausted)
-            // can always be set on the wrapping activity — even when no retry
-            // happens at all.
-            //
-            // `httpSendCount` counts the number of HTTP sends actually
-            // attempted (starting at 1 — we always make at least one). This
-            // is distinct from `attemptCount`, which counts retryable
-            // failures (transport errors + retryable status codes); they
-            // diverge on the exhaustion path where the failure count is
-            // recorded but no further send is made.
-            int httpSendCount = 0;
-            int retryEventNumber = 0;
-            bool retryExhausted = false;
-
             // Use TraceActivityAsync to wrap the entire retry logic
             return await this.TraceActivityAsync(async activity =>
             {
-                try
+                do
                 {
-                    do
+                    // Set the content for each attempt (if needed)
+                    if (requestContentClone != null && request.Content == null)
                     {
-                        // Set the content for each attempt (if needed)
-                        if (requestContentClone != null && request.Content == null)
-                        {
-                            request.Content = await CloneHttpContentAsync(requestContentClone);
-                        }
+                        request.Content = await CloneHttpContentAsync(requestContentClone);
+                    }
 
-                        // Try to send the request, catching transport-level errors for retry.
-                        // Use a per-request timeout to detect dead connections (e.g., TCP half-open).
-                        // This is separate from the caller's cancellation token (query timeout).
-                        using var requestTimeoutCts = _httpRequestTimeoutSeconds > 0
-                            ? new CancellationTokenSource(TimeSpan.FromSeconds(_httpRequestTimeoutSeconds))
-                            : null;
-                        using var linkedCts = requestTimeoutCts != null
-                            ? CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, requestTimeoutCts.Token)
-                            : null;
-                        var requestToken = linkedCts?.Token ?? cancellationToken;
+                    // Try to send the request, catching transport-level errors for retry.
+                    // Use a per-request timeout to detect dead connections (e.g., TCP half-open).
+                    // This is separate from the caller's cancellation token (query timeout).
+                    using var requestTimeoutCts = _httpRequestTimeoutSeconds > 0
+                        ? new CancellationTokenSource(TimeSpan.FromSeconds(_httpRequestTimeoutSeconds))
+                        : null;
+                    using var linkedCts = requestTimeoutCts != null
+                        ? CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, requestTimeoutCts.Token)
+                        : null;
+                    var requestToken = linkedCts?.Token ?? cancellationToken;
 
-                        httpSendCount++;
-                        try
-                        {
-                            response = await base.SendAsync(request, requestToken);
-                        }
-                        catch (Exception ex) when (!cancellationToken.IsCancellationRequested
-                            && _transportErrorRetryEnabled
-                            && IsTransientTransportException(ex, cancellationToken))
-                        {
-
-                            attemptCount++;
-                            lastTransportException = ex;
-
-                            activity?.SetTag("http.retry.attempt", attemptCount);
-                            activity?.SetTag("http.retry.transport_error", ex.GetType().Name);
-                            activity?.SetTag("http.retry.transport_error_message", ex.Message);
-
-                            int transportWaitSeconds = CalculateBackoffWithJitter(currentBackoffSeconds);
-                            lastErrorMessage = $"Transport error ({ex.GetType().Name}: {ex.Message}). Using exponential backoff of {transportWaitSeconds} seconds. Attempt {attemptCount}.";
-
-                            // Reset the request content for the next attempt
-                            request.Content = null;
-
-                            // Check if we would exceed the transport error retry timeout
-                            if (_retryTimeoutSeconds > 0 && totalTransportErrorRetrySeconds + transportWaitSeconds > _retryTimeoutSeconds)
-                            {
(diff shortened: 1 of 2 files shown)


csharp/src/RetryHttpHandler.cs
```diff
@@ -108,171 +109,243 @@ protected override async Task<HttpResponseMessage> SendAsync(
...
+                                break;
+                            }
+                            totalTransportErrorRetrySeconds += transportWaitSeconds;
+
+                            // Issue #479: emit a per-retry event BEFORE we
+                            // sleep so the trace makes the retry visible even
+                            // if the process dies during the delay. The
+                            // attempt_number is the 1-indexed retry counter
+                            // (1 = first retry).
+                            retryEventNumber++;
```
eric-wang-1990: Why not just use attemptCount?

csharp/src/RetryHttpHandler.cs
```diff
@@ -108,171 +109,243 @@ protected override async Task<HttpResponseMessage> SendAsync(
...
                 }
-
-                throw new DatabricksException(lastErrorMessage ?? "Service temporarily unavailable and retry timeout exceeded", AdbcStatusCode.IOError)
-                    .SetSqlState("08001");
+                finally
+                {
+                    // Issue #479: ALWAYS emit summary tags — even on the
+                    // no-retry-needed path, even when we threw. Without
+                    // always-set tags, dashboards have to detect "tag
+                    // absent" to distinguish "successful first try" from
```
eric-wang-1990: Note this will be mostly used for local user logs, so dashboard view is less important, re-evaluate if we need this PR

csharp/src/RetryHttpHandler.cs
eric-wang-1990: Fixed in dc6cd7a. Dropped `httpSendCount` / `retryEventNumber` entirely and used `attemptCount` directly as `attempt_number` in the `retry.attempt` events. `attemptCount` is already incremented before each retry event fires, so the per-retry numbering is identical (1 for first retry, 2 for second, ...).

---

csharp/src/RetryHttpHandler.cs
eric-wang-1990: Agreed — reframed the PR for local-log readers in dc6cd7a. Kept the per-retry `retry.attempt` events (the actually-useful diagnostic surface for someone reading their own logs) and dropped the always-fires summary tags `retry.total_attempts` / `retry.exhausted` along with the `try/finally` block that hosted them, since the dashboard-aggregation justification doesn't apply. The two summary-tag-only tests were deleted; the two event-asserting tests stay (with summary-tag assertions trimmed). Net diff is 173 insertions / 291 deletions, so the PR is significantly smaller and more focused now.

---

csharp/src/RetryHttpHandler.cs
```diff
@@ -161,6 +162,17 @@ protected override async Task<HttpResponseMessage> SendAsync(
                         }
                         totalTransportErrorRetrySeconds += transportWaitSeconds;
 
+                        // Issue #479: emit a per-retry event BEFORE we sleep so the
+                        // trace makes the retry visible even if the process dies
+                        // during the delay. attempt_number reuses attemptCount
+                        // (1-indexed retry counter).
+                        activity?.AddEvent("retry.attempt", new List<KeyValuePair<string, object?>>
+                        {
+                            new("attempt_number", attemptCount),
```
eric-wang-1990: We already log the retry attempt at line 146, why logging again here?

eric-wang-1990: Good catch — folded the per-retry detail into the event and dropped the overwritten per-retry tags (f566718), so the attempt is logged once.

- Transport path: removed `http.retry.attempt` / `http.retry.transport_error` / `http.retry.transport_error_message`; the `retry.attempt` event now carries `error_message` (the error type is already in `reason`).
- HTTP-status path: removed `http.retry.attempt` / `http.response.status_code`; the event now carries `status_code`.
- The terminal `http.retry.outcome` / `http.retry.total_attempts` tags stay — they fire once on exit, not per retry.

Added a transport-error event test (asserts reason + error_message) and a status_code assertion to the Retry-After test; all 29 RetryHttpHandler tests pass.

msrathore-db (approved):
Verified: 29/29 `RetryHttpHandlerTest` pass, including the new `retry.attempt` event tests. The event vocabulary is clean. LGTM.

One note worth surfacing: this also **removes** the pre-existing per-attempt span tags (`http.retry.attempt`, `http.response.status_code`, `http.retry.transport_error`, `http.retry.transport_error_message`) in favor of the events. The description frames it as a pure addition, but it's effectively a tag→event migration — if any dashboard/alert keys off those removed tags it would need updating. Fine by me given they were overwritten-per-retry anyway, just flagging it.

Issue #481, opened 2026-05-27
tracing(csharp): no semantic error classification (db.error.kind) on failure spans


eric-wang-1990:
**Severity:** Low

## Observation
When a query fails, the only error classification is `exception.type` (often `TaskCanceledException` or `TTransportException`). These exception types are ambiguous between several failure modes:

- `TaskCanceledException`: client-side timeout? user-cancel via `Cancel()`? mid-flight network drop?
- `TTransportException`: server-side cancel? network drop? Thrift framing error?

A debugger can read the `exception.message` and tell the difference, but automated dashboards / alerts can't group-by an error kind.

## Suggested fix
Add a `db.error.kind` (or `error.category`) tag with values like `query_timeout`, `network`, `user_cancel`, `server_error`, `auth_failed` on any span that ends with `Status=Error`. The driver already knows the kind internally (it set the timeout, it called Cancel, etc.) — just surface it.

## Evidence
- `Traces.StatementTests.20260527_144246/` — 93 `Status=Error` spans, all classify only via `exception.type` (`TaskCanceledException` / `TTransportException` / `IOException`)
- `Traces.DriverTests.20260527_153928/` — `Status=Error` spans for invalid-server use `IOException` ("nodename nor servname...") and invalid-auth uses HTTP-401-derived exceptions; no `error.kind` to group on

PR #499, opened 2026-05-28
tracing(csharp): tag failures with db.error.kind

eric-wang-1990:
## Motivation

When a Databricks query fails, the only error classification on the
resulting Status=Error spans is `exception.type` (typically
`TaskCanceledException` or `TTransportException`). That exception type
is ambiguous between several distinct failure modes:

- `TaskCanceledException` — client-imposed `QueryTimeoutSeconds`? user-
  initiated `Cancel()`? mid-flight network drop?
- `TTransportException` — server-side cancel? network drop? Thrift
  framing error?

Automated dashboards and alerts can't group-by an error kind without
parsing `exception.message` strings. The driver knows internally which
kind it is (it set the timeout token; it received the cancel; it caught
the transport error) — this PR surfaces that knowledge as a stable
`db.error.kind` enum-string tag on Status=Error spans.

See issue #481.

## Taxonomy

The full taxonomy proposed in #481:

| Kind             | Implemented? | Notes                                        |
|------------------|--------------|----------------------------------------------|
| `query_timeout`  | YES          | Client-side `QueryTimeoutSeconds` fired      |
| `user_cancel`    | YES          | Caller invoked `Cancel()` on the statement   |
| `network`        | DEFERRED     | Needs separate exception-type analysis       |
| `server_error`   | DEFERRED     | Needs Thrift TStatus inspection (submodule)  |
| `auth_failed`    | DEFERRED     | Needs 401/403 catch site routing             |
| `protocol_error` | DEFERRED     | Needs Thrift framing / Arrow decode signals  |

The four deferred kinds either need catch-site changes in the
hiveserver2 submodule (`server_error` lives where `TStatus != SUCCESS`
is inspected; `protocol_error` lives in the Thrift transport layer)
or a follow-up exception-type taxonomy patch on the parent listener.
Per the issue and TDD plan: better to land two classifications
cleanly than to fudge all six. Issue #481 stays open after this merge
for the remaining four.

### `csharp/src/DatabricksStatement.cs`

1. Registers a process-global `ActivityListener` in an eagerly-initialized
   static field (with an explicit `static DatabricksStatement()`
   constructor to defeat `beforeFieldInit` and guarantee the listener
   is in place before the first execute span emits).
2. The listener requests `PropagationData` sampling — the lowest level
   that still delivers `ActivityStopped` callbacks — so it never
   degrades exporters that subscribe at `AllDataAndRecorded` (effective
   recording level across all listeners is the max).
3. On `ActivityStopped`, it inspects only Status=Error spans whose
   `OperationName` ends with `.ExecuteQueryAsyncInternal`,
   `.ExecuteUpdateAsyncInternal`, `.ExecuteStatementAsync`, or
   `.PollForResponseAsync` (the spans listed in the evidence trace
   `Traces.StatementTests.20260527_144246/`). Other spans are
   untouched.
4. If the exception event's `exception.type` is in the cancellation
   class — `TaskCanceledException`, `OperationCanceledException`,
   `TimeoutException`, or `TTransportException` — the listener tags
   the span with `db.error.kind`. The value is chosen by an
   `AsyncLocal<ErrorKindContext>` populated by the parent's Execute*
   overrides; the parent's `Cancel()` override flips a flag on the
   context. The AsyncLocal flows into the submodule's
   `TraceActivityAsync` delegate, so the listener sees the right
   context when the inner activity stops.
5. Spans with `exception.type` outside the cancellation class are
   left unset on purpose — per the issue, conservatively leave
   `db.error.kind` UNSET rather than guess (`network` /
   `server_error` / `protocol_error` etc. need their own
   classification paths).
6. Tag mutation during `ActivityStopped` is documented safe:
   `Activity.IsStopped` flips to true only after every registered
   `ActivityStopped` callback returns.

The hiveserver2 submodule is read-only from this repo, so this
listener-plus-AsyncLocal pattern is the least invasive in-driver fix
that targets the very spans the issue calls out. It follows the
established pattern from the merged
`db.response.returned_rows = -1` cleanup (PR #491).

### `csharp/test/E2E/StatementTests.cs`

Adds two E2E tests:

- `StatementTimeout_TagsErrorKindAsQueryTimeout_Issue481` — sets
  `QueryTimeoutSeconds=5` against the existing
  `LongRunningStatementTimeoutTestData.LongRunningQuery`, asserts the
  expected `TimeoutException`, and then asserts that at least one
  captured Status=Error span carries `db.error.kind="query_timeout"`.
- `StatementCancel_TagsErrorKindAsUserCancel_Issue481` — sets a
  generous `QueryTimeoutSeconds=120` so the timeout cannot fire, runs
  the same long-running query, calls `statement.Cancel()` from a
  background task after a short delay, and asserts that at least one
  Status=Error span carries `db.error.kind="user_cancel"` AND that no
  span is mislabeled as `query_timeout`.

Both tests subscribe an `ActivityListener` scoped to
`AdbcDrivers.*` sources to capture spans for assertion.

## TDD evidence

Branch commits (chronological):

1. `c746719 test(csharp): assert db.error.kind on timeout failure (failing for #481)` (RED)
   — Test runs and FAILS with `Inspected 3 Status=Error span(s); none
   carry db.error.kind`. Spans observed: `SendAsync`,
   `HiveServer2Statement.ExecuteStatementAsync`,
   `HiveServer2Statement.ExecuteQueryAsyncInternal`.
2. `ef94386 fix(csharp): tag query_timeout failures with db.error.kind for #481` (GREEN)
   — Listener registered + Execute* overrides install AsyncLocal +
   Cancel() flips user-cancel flag. Test now PASSES.
3. `f8831c2 test(csharp): assert db.error.kind=user_cancel distinguishes Cancel from timeout (#481)`
   — Regression test for the user_cancel branch of the same
   classifier. Passes against the same fix.

## Regression check

- `StatementTests` class: 96 passed / 3 skipped / 0 failed against the
  configured Thrift workspace.
- `DriverResilienceTests` class: 7 passed / 0 failed.
- `Unit` tests: 846 passed / 0 failed.

Refs #481

### DIFF
--- a/csharp/src/DatabricksStatement.cs
+++ b/csharp/src/DatabricksStatement.cs
@@ -24,6 +24,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


eric-wang-1990:
Superseded — closing per same rationale as PR #491. The process-global ActivityListener + AsyncLocal pattern is heavier than the problem warrants. The proper fix is in the hiveserver2 submodule where the catch sites are. A new submodule PR will land error.kind classification at the inner catch sites, mirroring the playbook used for #484 → adbc-drivers/hiveserver2#57. Will reopen / file follow-up parent PR after the submodule fix merges (to bump the submodule pointer + add the parent regression test).

PR #500, opened 2026-05-30
feat(csharp): enable feature flag cache by default (hardened)

eric-wang-1990:
## Summary

**Enables the feature flag cache by default** (`adbc.databricks.feature_flag_cache_enabled` now defaults to `true`; only an explicit, parseable `"false"` disables it) and hardens the cache so default-on is safe. Healthy-path behavior is unchanged.

### Hardening
1. **Default timeout 10s → 5s** (`HttpClientFactory.CreateFeatureFlagHttpClient`). The feature-flag fetch is on the (blocking, sync-over-async) connection-open path, so this halves the worst-case stall. The `adbc.databricks.feature_flag_timeout_seconds` override still wins.

2. **Negative caching.** A failed fetch (timeout / 429 / 5xx) now records a `Failed` status, and `FeatureFlagCache` caches that context with a short **fixed absolute** TTL (60s, `FeatureFlagContext.DefaultNegativeTtl`) instead of throwing it away. Previously a timeout cached nothing, so every cold connection re-fetched and re-stalled (N × timeout, N endpoint calls). Now it's **one fetch per host per 60s window**; concurrent connections to the same workspace serialize on the existing create-lock and then hit the cache, so they all see a **consistent** result.

3. **Timeout vs. cancellation.** An HttpClient timeout surfaces as `TaskCanceledException` with the request token *not* cancelled. We now treat that as a failed fetch (negatively cached) while still propagating genuine caller/dispose cancellation. This also fixes a latent bug where a timed-out **background** refresh would hit `catch (OperationCanceledException) { break; }` and permanently kill the refresh loop.

### Behavior change to call out
With the default flipped, a cold connection now makes a best-effort call to the connector-service feature-flags endpoint by default. It's bounded (5s timeout, negatively cached on failure) and **failures are swallowed** — a down/slow endpoint never breaks `Connect`, it just means flags aren't applied that window. Set `adbc.databricks.feature_flag_cache_enabled=false` to opt out.

### Unchanged behavior
- Healthy contexts: sliding cache entry + background refresh at the server `ttl_seconds`.
- A healthy context whose background refresh later fails retains its last-known-good flags and retries at the positive cadence.
- Connection-open fetch stays **blocking** on purpose, for consistent flags across connections to the same workspace.

## Behavior matrix

| Path | Recovery |
|---|---|
| Healthy | background loop at server `ttl_seconds`, sliding cache entry |
| Cold / initial-fetch failure | fixed 60s absolute negative entry → recreation on next connection; 1 fetch/host/window |
| Healthy → later background failure | retain last-good flags, retry at positive cadence |

## Tests
**Unit** (`FeatureFlagCacheTests`):
- 5s default timeout + `feature_flag_timeout_seconds` override honored.
- Negative caching: 429 → fixed 60s; 500 → 60s; timeout → 60s; 200 → `Healthy`, no negative TTL.
- Timeout does not throw out of `CreateAsync`; pre-cancelled caller token propagates.
- Default-enabled decision: absent/unparseable → enabled; explicit `false` → disabled (returns local unchanged).

**E2E** (`FeatureFlagCacheE2ETest`):
- Asserts the connector-service endpoint returns a **non-empty** flag set (happy path).
- The helper no longer sets the flag, so it exercises the **default-on** path. Verified against a live workspace — **9 flags** returned via the default, `ttl_seconds=3600`.
- `TestFeatureFlagCache_SingleExternalCallAcrossConnections` opens 5 connections to the same host and counts actual fetches via an ActivityListener on the feature-flag source. Verified live: **5 connections → 1 external connector-service call** (rest served from the shared singleton cache).

Local run: unit `FeatureFlagCacheTests` 31 passed / 0 failed; full `FeatureFlagCacheE2ETest` class (5 tests) passed live.

### DIFF
--- a/csharp/src/DatabricksDatabase.cs
+++ b/csharp/src/DatabricksDatabase.cs
@@ -24,6 +24,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 using AdbcDrivers.Databricks.StatementExecution;
 using Apache.Arrow.Adbc;
 using AdbcDrivers.HiveServer2;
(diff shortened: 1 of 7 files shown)


msrathore-db (approved):
Verified live: 31 unit + all 5 E2E tests pass. The default-on path fetched 9 flags (`ttl_seconds=3600`), and the 5-connections→1-external-call assertion held (shared singleton cache) — matches the reported numbers. The hardening is solid: negative caching (60s absolute), the `when (cancellationToken.IsCancellationRequested)` split for timeout-vs-cancellation, and no background loop on a failed initial fetch. LGTM.

Minor (non-blocking): `FeatureFlagContext.NegativeTtl` looks unused — the cache uses `DefaultNegativeTtl` directly.

jadewang-db (approved):

PR #505, opened 2026-06-02
ci: use UC REST API for schema create/drop in Reyden nightly

eric-wang-1990:
## Summary

- Fixes the Reyden REST nightly run (introduced in #504) which failed immediately on `Provision per-run schema` with `[USER_ERROR] Unsupported CREATE type: SCHEMA`
- Reyden warehouse rejects DDL via the SQL statement API; switch to the Unity Catalog REST API (`/api/2.1/unity-catalog/schemas`) for both schema creation and deletion
- Seed step is unchanged — it only runs `CREATE TABLE` / `INSERT` DML which Reyden does support

## Test plan

- [ ] Trigger `reyden-rest-nightly` via `workflow_dispatch` after merge to verify schema is created and tests run

Fixes first-run failure from #504.

---
_This PR was created with [GitHub MCP](http://go/mcps)._


eric-wang-1990:
Closing as stale — conflicting (DIRTY) reyden REST-nightly CI fix. Reopen/rebase if the reyden REST nightly still needs it.

PR #509, opened 2026-06-03
feat(csharp): make the driver AOT/trim-safe under net10.0

CurtHagenlocher:
## What's Changed

Adds net10.0 to the driver's target frameworks and enables the trim, single-file, and NativeAOT Roslyn analyzers for it (IsAotCompatible).

A normal `dotnet build` now surfaces IL2xxx/IL3xxx warnings — which, with TreatWarningsAsErrors, fail the build — so the driver is verified AOT-ready without yet producing a native build.

Enabling the analyzers flagged one category of problem: reflection-based System.Text.Json. Fixes:
  - Route the concrete-DTO (de)serialization through source-generated JsonSerializerContexts: SEA request/response models, the DESC-EXTENDED result, the feature-flags response, the config dictionary, and the telemetry request envelope. Options mirror the prior runtime JsonSerializerOptions; unused options fields removed.
  - ComplexTypeSerializingStream serializes polymorphic Arrow value graphs (ARRAY/MAP/STRUCT), which source-gen can't model. Replace JsonSerializer.Serialize(object) with a recursive Utf8JsonWriter walk over the closed set of Arrow scalar/list/dict types — same output, no reflection.
  - TelemetryFrontendLog nests a protobuf type whose graph trips the source generator (SYSLIB1031: nested `Type` enum name collisions), even with a property-level [JsonConverter]. Serialize its small, stable envelope by hand (TelemetryPayloadWriter), reusing the proto converter for the leaf. The proto-free TelemetryRequest stays on source-gen; the reflection-based TelemetryJsonOptions is kept for the test suite.

All four TFMs (netstandard2.0;net472;net8.0;net10.0) build clean with no IL/SYSLIB warnings; net10.0 goes from 28 IL warnings to 0. The 825 unit tests pass, including the telemetry tests that exercise the production serialization path.


davidhcoe (approved):

eric-wang-1990:
Thanks @CurtHagenlocher , may I know what is the reason for this PR?
I think PowerBI still use net472, why do we need to enable dotNet 10?
Is this related to the perf improvement work?

My only concern is since it does not use reflection anymore, any new fields will need an explicitly map.

csharp/src/ComplexTypeSerializingStream.cs
```diff
@@ -139,16 +142,131 @@ private RecordBatch ConvertComplexColumns(RecordBatch batch)
...
+                case IEnumerable<object?> list:
+                    writer.WriteStartArray();
+                    foreach (object? item in list)
+                        WriteJsonValue(writer, item);
+                    writer.WriteEndArray();
+                    break;
+                default:
+                    // Rare Arrow scalar types not enumerated above (e.g. native interval structs):
+                    // fall back to their invariant string form to keep the output valid JSON.
+                    writer.WriteStringValue(Convert.ToString(value, CultureInfo.InvariantCulture));
```
eric-wang-1990: The `default:` arm silently flips category — anything that should be a JSON number (e.g. a future boxed `Int128`/`UInt128`/`Half` or another Arrow scalar not enumerated above) lands as a JSON string instead. The prior `JsonSerializer.Serialize(object)` would have picked the right converter via reflection, so this is a behavior change for types the switch hasn't anticipated, not just for the documented "rare native interval struct" case.

Two low-cost defenses worth considering:
- A `Debug.Assert(false, $"Unhandled Arrow scalar type {value?.GetType()}")` in the default arm so existing unit tests catch newly-added types in DEBUG builds before they ship.
- Or, explicit cases for the integer/float scalars Arrow may add over time (cheap to enumerate, removes the ambiguity for numbers).

Not blocking — the fallback keeps output valid JSON — just a soft place where the trim-safe rewrite quietly changes semantics vs. the reflection path.

---

csharp/src/Telemetry/TelemetryPayloadWriter.cs
```diff
@@ -0,0 +1,113 @@
...
+    /// </para>
+    ///
+    /// <para>
+    /// Output matches the reflection serializer configured by <see cref="TelemetryJsonOptions"/>:
+    /// the default encoder, explicit snake_case property names, and null properties omitted.
+    /// </para>
+    /// </summary>
+    internal static class TelemetryPayloadWriter
+    {
+        internal static string SerializeFrontendLog(TelemetryFrontendLog log)
```
eric-wang-1990, line 47: This hand-written serializer is fragile against schema additions. If anyone adds a property to `TelemetryFrontendLog`, `FrontendLogContext`, or `FrontendLogEntry`, the reflection path would have emitted it automatically; this writer silently omits it and the wire payload ships with a missing field without any test failing.

A cheap defense is a round-trip test that serializes a fully-populated `TelemetryFrontendLog` both via the existing reflection-based `TelemetryJsonOptions` and via `TelemetryPayloadWriter.SerializeFrontendLog`, and asserts the two strings are equal. Drift between the two paths then becomes a test failure rather than a silent prod regression. Alternatively, a reflection-based test that enumerates declared properties of those three types and asserts each appears in the writer's output for a populated instance.

(The reflection-based `TelemetryJsonOptions` is already kept around for the test suite per the PR description, so wiring up the round-trip should be straightforward.)

---

csharp/src/Telemetry/TelemetryPayloadWriter.cs
```diff
@@ -0,0 +1,113 @@
...
+    /// source generator walks that graph during type discovery and fails with SYSLIB1031 (name
+    /// collision) even when the proto member carries a custom <c>[JsonConverter]</c>. So rather
+    /// than route this type through a <see cref="JsonSerializerContext"/>, we write its small,
+    /// stable envelope by hand with <see cref="Utf8JsonWriter"/> and delegate the proto leaf to
+    /// <see cref="OssSqlDriverTelemetryLogJsonConverter.WriteValue"/> — the same formatter the
+    /// reflection-based path uses.
+    /// </para>
+    ///
+    /// <para>
+    /// Output matches the reflection serializer configured by <see cref="TelemetryJsonOptions"/>:
```
eric-wang-1990, line 39: Minor doc inconsistency worth resolving: this file says output uses "explicit snake_case property names" while `TelemetryJsonContext`'s docstring says it mirrors "camelCase property names". Both are correct in context — the `TelemetryRequest` envelope uses camelCase via `PropertyNamingPolicy`, while `TelemetryFrontendLog` and its nested types use snake_case because the proto-generated models carry explicit `[JsonPropertyName("snake_case")]` attributes that override the policy — but reading the two docstrings side by side leaves a reader thinking the two paths disagree.

One extra sentence here clarifying "the snake_case names come from `[JsonPropertyName]` attributes on the proto-generated models, not from a global naming policy" would close the loop.

---

CurtHagenlocher:
> Thanks @CurtHagenlocher , may I know what is the reason for this PR? I think PowerBI still use net472, why do we need to enable dotNet 10? Is this related to the perf improvement work?

The only reason to enable `net10.0` is its improved support for AOT compilation. AOT compilation means that we can compile the driver to native code and load it as if it were a native driver through the C API instead of the managed API. There are pros and cons to doing this, but the cons are probably comparable to loading a Go-based driver the same way. The advantages are that we get some net10.0 perf improvements even when running under net472, and that it results in greater isolation with the PBI/PQ dependencies -- potentially making it easier to have two of them side-by-side.

For now, this is an option we're considering rather than something we've decided to definitely do.

> My only concern is since it does not use reflection anymore, any new fields will need an explicitly map.

I believe that particular serializer is for converting structured Arrow data into JSON. We can probably gain confidence and efficiency by going straight from Arrow into the Json writer instead of using .NET structures as an intermediate. I'll look into making that change.

csharp/src/ComplexTypeSerializingStream.cs
CurtHagenlocher: Rewritten to work differently.

CurtHagenlocher:
I've rewritten the structured types JSON serializer to omit the intermediate .NET serialization and added new tests for it. I think the new version is more verifiably complete.

csharp/src/ComplexTypeSerializingStream.cs
```diff
@@ -175,53 +191,123 @@ private static HashSet<int> DetectComplexColumns(Schema schema)
...
+#endif
+                case ArrowTypeId.Int8: writer.WriteNumberValue(((Int8Array)array).GetValue(index)!.Value); break;
+                case ArrowTypeId.Int16: writer.WriteNumberValue(((Int16Array)array).GetValue(index)!.Value); break;
+                case ArrowTypeId.Int32: writer.WriteNumberValue(((Int32Array)array).GetValue(index)!.Value); break;
+                case ArrowTypeId.Int64: writer.WriteNumberValue(((Int64Array)array).GetValue(index)!.Value); break;
+                case ArrowTypeId.String: writer.WriteStringValue(((StringArray)array).GetString(index)); break;
+                case ArrowTypeId.LargeString: writer.WriteStringValue(((LargeStringArray)array).GetString(index)); break;
+#if NET6_0_OR_GREATER
+                case ArrowTypeId.Time32: writer.WriteRawValue(((Time32Array)array).GetTime(index)!.Value.ToString("HH:mm:ss.ffffff")); break;
+                case ArrowTypeId.Time64: writer.WriteRawValue(((Time64Array)array).GetTime(index)!.Value.ToString("HH:mm:ss.ffffff")); break;
```
eric-wang-1990: **`WriteRawValue` on time strings produces invalid JSON and will throw at runtime.**

On .NET 6+, `Time32Array.GetTime()` / `Time64Array.GetTime()` return a `TimeOnly?`, and `.ToString("HH:mm:ss.ffffff")` produces a string like `12:30:45.123456`. `Utf8JsonWriter.WriteRawValue` expects a syntactically valid JSON token (number, string, boolean, null, object, or array). An unquoted time string with colons is not valid JSON, so `WriteRawValue` will throw a `JsonException` at runtime for any Time32/Time64 value nested inside a complex type (ARRAY, MAP, or STRUCT).

The non-.NET-6 fallback path just below correctly uses `WriteStringValue`, confirming the intent is a quoted string.

```suggestion
                case ArrowTypeId.Time32: writer.WriteStringValue(((Time32Array)array).GetTime(index)!.Value.ToString("HH:mm:ss.ffffff")); break;
                case ArrowTypeId.Time64: writer.WriteStringValue(((Time64Array)array).GetTime(index)!.Value.ToString("HH:mm:ss.ffffff")); break;
```

CurtHagenlocher: I'll add a test case for this later.

sizzles:
Just to add .NET 10 compatibility would be helpful. 

I have a use case of needing to read data from Databricks, not via SQL endpoint, but via a Windows Azure Web App. 

In which you can't easily install ODBC drivers (as startup commands are only available for Linux containers_. 

This solution has been working really well for us so far in proof of concept testing, but we would need to upgrade our solution from .NET 8 at some point.

CurtHagenlocher:
> Just to add .NET 10 compatibility would be helpful.

Most net8.0 packages should be fully compatible with net10.0; have you tried just upgrading your build and seeing what happens?

sizzles:
> > Just to add .NET 10 compatibility would be helpful.
> 
> Most net8.0 packages should be fully compatible with net10.0; have you tried just upgrading your build and seeing what happens?

Thanks - will give that a try!

eric-wang-1990:
**Map keys fall through to locale-dependent `ValueAt().ToString()` (~L328)** — a `MAP<DATE,…>` key renders `"6/15/2024 12:00:00 AM"` instead of `"2024-06-15"`. Reuse the value-path formatting so keys stay deterministic.

Nits: nested DayTime interval (~L222) emits non-canonical `"3.12:30:45.1230000"` (test cements it); key ordering now matches server source-order — good.

PR #510, opened 2026-06-04
fix(csharp): surface closed/expired sessions as a clear, typed error

jadewang-db:
When a server-side session is closed or times out due to inactivity, the next Thrift call returns HTTP 400 "Invalid SessionHandle". The driver previously surfaced this as a misleading "An unexpected error occurred while fetching results / Couldn't connect to server" wrapper, and kept reusing the now-stale handle so every subsequent call failed the same opaque way until the caller disposed and re-created the connection.

Tier 1 (clear, typed error): detect the closed/expired-session signature and surface it as a typed DatabricksSessionExpiredException (AdbcStatusCode.InvalidState) with a clear message, preserving the original error as InnerException for diagnostics.

Tier 2 (fail fast): mark the connection's session invalid on first detection so subsequent statements fail fast with the same typed error instead of reusing the dead handle.

Detection walks the inner/aggregate exception chain for the stable server phrase "Invalid SessionHandle" (covers both the inactivity-timeout "Session [..] is closed" and explicit-close variants) and is careful not to misclassify connectivity errors or invalid *operation* handles.

Tests:
- Unit tests for the classifier (no workspace required, runs in CI).
- E2E test that reproduces the condition on demand by closing the session out-of-band, asserting both the typed Tier 1 error and Tier 2 fast-fail.

### DIFF
--- a/csharp/src/DatabricksSessionExpiredException.cs
+++ b/csharp/src/DatabricksSessionExpiredException.cs
@@ -1,0 +1,93 @@
+/*
+* Copyright (c) 2025 ADBC Drivers Contributors
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+using System;
+using Apache.Arrow.Adbc;
+
+namespace AdbcDrivers.Databricks
+{
+    /// <summary>
+    /// Raised when the server-side session backing a connection has been closed or has
+    /// expired (e.g. due to an inactivity timeout) and the connection can no longer be used.
+    ///
+    /// The server reports this as HTTP 400 with a Thrift error message containing
+    /// "Invalid SessionHandle". Without this typed exception the condition surfaces as a
+    /// generic, misleading "An unexpected error occurred while fetching results / Couldn't
+    /// connect to server" error, which makes it impossible for callers to distinguish a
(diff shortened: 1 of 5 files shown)


csharp/src/DatabricksSessionExpiredException.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// </summary>
+    public class DatabricksSessionExpiredException : DatabricksException
+    {
+        /// <summary>
+        /// Substring present in the server's error message for a closed/expired session.
+        /// Both the inactivity-timeout variant ("Invalid SessionHandle: Session [..] is closed")
+        /// and the explicitly-closed variant ("Invalid SessionHandle: SessionHandle [..]")
+        /// contain this phrase.
+        /// </summary>
+        internal const string ServerErrorSignature = "Invalid SessionHandle";
```
msrathore-db, line 42: is there no better way to identify this?

csharp/src/DatabricksStatement.cs
```diff
@@ -211,60 +211,118 @@ private void RecordError(StatementTelemetryContext ctx, Exception ex)
             CaptureRetryCount(ctx);
         }
 
-        public override QueryResult ExecuteQuery()
+        /// <summary>
+        /// Tier 2 fast-fail: if the connection's server-side session is already known to be
```
msrathore-db, line 215: Tier 2 fast-fails emit no telemetry

msrathore-db (commented):
We need to do the same for metadata and fetch paths also right?

eric-wang-1990:
I think this logic should happen in the HiveServer2Connection class: https://github.com/adbc-drivers/hiveserver2/blob/main/csharp/src/AdbcDrivers.HiveServer2/Hive2/HiveServer2Statement.cs#L360C85-L360C105.

We can change the SessionHandle to something like this:
```
internal TSessionHandle SessionHandle
{
    get
    {
        if (_isSessionInvalidated)
        {
            throw new AdbcException(
                "The session was closed by the server.");
        }

        return _sessionHandle
            ?? throw new AdbcException(
                "Connection is not open.");
    }

    private set => _sessionHandle = value;
}
```
Then this will guard for every Thrift operation and we do not need to manually edit the callsite  one by one.
WDYT?

PR #511, opened 2026-06-04
test(csharp): skip Thrift-only E2E tests on the SEA protocol

eric-wang-1990:
## What

`SeaMetadataE2ETests` hardcoded the protocol per test (`Thrift_*`/`SEA_*` pairs duplicating assertions against forced connections, plus `*_ThriftAndSEA_*` tests comparing protocols in-process). On the SEA-only Reyden nightly the forced-Thrift sessions 404 (`ENDPOINT_NOT_FOUND`, no Thrift endpoint).

This makes the metadata tests **protocol-agnostic**: each runs on whatever protocol the suite is configured with (from the driver parameters / config via `TestConfiguration`) and asserts a fixed expected result. Both protocols are covered by the CI matrix (Thrift CI + SEA/Reyden nightly), without comparing protocols inside a test.

> Note: the CloseOperation/CloudFetch protocol changes that were originally in this branch are now handled by #508 (merged). After merging `main`, this PR's only effective change is `SeaMetadataE2ETests.cs`.

## Changes (`SeaMetadataE2ETests`)

- Removed the `Thrift_*`/`SEA_*` duplication and the `*_ThriftAndSEA_*` in-process comparisons; removed the `CreateThriftConnection`/`CreateSeaConnection` helpers. A single `CreateConnection()` uses the configured protocol.
- `GetColumnsExtended_ReturnsExpectedColumns` asserts the protocol-invariant column metadata (`COLUMN_NAME`, `DATA_TYPE`, `BASE_TYPE_NAME`, `COLUMN_SIZE`, `DECIMAL_DIGITS`) against the shared golden file `Resources/result_get_column_extended_all_types.json` + the 32-column schema — preserving the old cross-protocol coverage as concrete values. `BUFFER_LENGTH` excluded (Thrift Int8 vs SEA Int32, PECO-3008).
- `GetColumnsExtended_FallbackAndDescTable_SameResults` runs on the configured protocol — `use_desc_table_extended` is honored on both Thrift and SEA, so fallback vs DESC TABLE EXTENDED must agree on either.
- `GetCatalogs` / `GetTables` / `GetPrimaryKeys` / `GetTableSchema` / `GetTableTypes` / `GetInfo` assert fixed expected results on the configured protocol.

## Skip behavior

`SeaMetadataE2ETests` has **no protocol skips** — every test runs on whatever protocol the suite is configured with. The only `Skip` is `SkipIfNotConfigured()` (no test config available; protocol-independent).

## Verification

Built locally and ran against a live warehouse over **both** protocols: `SeaMetadataE2ETests` 8/8 pass on SEA and 8/8 on Thrift (including `FallbackAndDescTable`).

## Out of scope

The cross-protocol parity tests in `IntervalValueTests` (`INTERVAL009/010`) still open both connections to compare; they need a separate "dual-protocol required" gate and are untouched here.


csharp/test/E2E/StatementExecution/SeaMetadataE2ETests.cs
```diff
@@ -72,6 +69,11 @@ private AdbcConnection CreateSeaConnection(Dictionary<string, string>? extraPara
             return db.Connect(new Dictionary<string, string>());
         }
 
+        private AdbcConnection CreateThriftConnection() => CreateConnection("thrift");
```
eric-wang-1990: You should not have these, the protocol is decided by the run argument.
You do not need to know which protocol it is, you just need to make sure the assert always assert same reuslts for Thrift and SEA or any protocol

eric-wang-1990: Done — removed `CreateThriftConnection`/`CreateSeaConnection` and the cross-protocol comparison tests entirely (pushed in `de61559`).

The metadata tests no longer know or pick the protocol: each uses a single `CreateConnection()` that takes whatever the run is configured with, and asserts a **fixed expected result**. Running across the CI matrix (Thrift CI + SEA/Reyden nightly) gives parity for free — no in-test protocol comparison.

- `Thrift_*`/`SEA_*` pairs and `*_ThriftAndSEA_*` comparisons → gone.
- `GetColumnsExtended_ReturnsExpectedColumns` asserts the concrete ordinal column list of `all_column_types` + the 32-column metadata schema.
- `GetColumnsExtended_FallbackAndDescTable_SameResults` compares two SEA *feature-flag* configs (not protocols) and skips unless the run is on SEA, since DESC TABLE EXTENDED is SEA-only.
- `GetTableTypes` / `GetInfo` run on the configured protocol.

eric-wang-1990: Follow-up (`a46bc9c`): restored the per-column coverage with concrete expected values so nothing is lost vs the old Thrift-vs-SEA comparison.

`GetColumnsExtended_ReturnsExpectedColumns` now asserts the result against the shared golden file `Resources/result_get_column_extended_all_types.json` (the same one `StatementTests.CanGetColumnsExtended` uses), checking the protocol-invariant columns per row — `COLUMN_NAME`, `DATA_TYPE`, `BASE_TYPE_NAME`, `COLUMN_SIZE`, `DECIMAL_DIGITS` — plus the 32-column metadata schema. These are concrete expected values that hold on any protocol, so the run argument still decides the protocol. `BUFFER_LENGTH` stays excluded (Thrift Int8 vs SEA Int32, PECO-3008).

One gap I couldn't fill: `GetTableSchema` (`cross_ref_customers`) has no DDL/golden in the repo, so it still only asserts the first field is `customer_id`. If you want full field-name/type assertions there, we'd need a golden for that table — happy to add one.

msrathore-db (approved):

PR #516, opened 2026-06-10
fix(csharp): report informational version in telemetry driver_version

eric-wang-1990:
## What

Telemetry's `driver_version` now reports the assembly's **`AssemblyInformationalVersion`** (e.g. `1.1.4-SNAPSHOT+<sha>`) instead of the bare numeric **`AssemblyVersion`** (`1.1.4`).

## Why

`DatabricksConnection.InitializeTelemetry` passed `s_assemblyVersion` into `ConnectionTelemetry.Create`, and `s_assemblyVersion = ApacheUtility.GetAssemblyVersion(...) = Assembly.GetName().Version.ToString(3)` — a purely numeric `major.minor.build`. That **drops the `-SNAPSHOT` prerelease label and the `+<commit-sha>` build metadata**, so:

- Every `1.1.4` build (nightly, PR CI, release — each a different commit) reports an *identical* `driver_version`; you can't pinpoint which build/commit emitted an event (defeats build-cutover detection).
- It's inconsistent with historical telemetry: production lumberjack rows show `0.23.0-SNAPSHOT+b716671<sha>` (an informational version), so the field's semantics had silently regressed to the coarser numeric form.
- It diverges from the JDBC driver's version format.
- It's fragile: `AssemblyVersion` is the strong-name binding identity and is commonly pinned; if it's ever pinned for binding stability, telemetry would report a stale/wrong version.

The assembly *selection* was already correct (the `new` override reads the Databricks assembly, not the base Apache/HiveServer2 one) — only the version *field* was wrong.

## Change

- Add `DatabricksConnection.s_telemetryDriverVersion` = `ApacheUtility.GetAssemblyProductVersion(typeof(DatabricksConnection), s_assemblyVersion)` (the informational version, falling back to the numeric version if the attribute is absent), and feed it into `ConnectionTelemetry.Create`.
- `s_assemblyVersion` (numeric) stays the source for the ADBC `AssemblyVersion` property and trace tags — unchanged.

## Test

`ConnectionTelemetryDriverVersionTests` pins that:
1. The telemetry version equals the assembly's `AssemblyInformationalVersion`.
2. When a `-SNAPSHOT`/`+sha` suffix is present (true for all dev/CI builds via `VersionSuffix=SNAPSHOT`), it does **not** collapse to the bare numeric `AssemblyVersion` — the discriminating regression guard (this assertion fails on the pre-fix numeric source).

## Verification

Could not build/run locally due to an unrelated `net10.0` submodule vs. SDK 9.0.304 mismatch (all build errors are in `hiveserver2/*`, none in the changed files). CI will compile and run `ConnectionTelemetryDriverVersionTests`.

### DIFF
--- a/csharp/src/DatabricksConnection.cs
+++ b/csharp/src/DatabricksConnection.cs
@@ -55,6 +55,14 @@ namespace AdbcDrivers.Databricks
         internal static new readonly string s_assemblyVersion = ApacheUtility.GetAssemblyVersion(typeof(DatabricksConnection));
         internal static readonly string DriverVersion = s_assemblyVersion;
 
+        // Version reported in the telemetry payload's driver_version. Uses the assembly's
+        // AssemblyInformationalVersion (e.g. "1.1.4-SNAPSHOT+<sha>") rather than the numeric
+        // AssemblyVersion in s_assemblyVersion (e.g. "1.1.4"), so telemetry can pinpoint the
+        // exact build/commit and matches the JDBC driver's version format. Falls back to the
+        // numeric version if the assembly carries no AssemblyInformationalVersionAttribute.
+        internal static readonly string s_telemetryDriverVersion =
+            ApacheUtility.GetAssemblyProductVersion(typeof(DatabricksConnection), s_assemblyVersion);
+
         /// <summary>
         /// The environment variable name that contains the path to the default Databricks configuration file.
         /// Takes precedence over <see cref="DefaultConfigEnvironmentVariable"/> when both are set.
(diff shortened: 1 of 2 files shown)


eric-wang-1990:
Closing — not a needed fix. The current code emits a correct `driver_version` (e.g. `1.1.4`); sourcing it from AssemblyVersion vs AssemblyInformationalVersion only differs for SNAPSHOT/dev/CI builds (`1.1.4` vs `1.1.4-SNAPSHOT+<sha>`), and is identical for clean releases. It's a build/commit-traceability enhancement, not a correctness bug, so closing for now. Can revisit if distinguishing SNAPSHOT/nightly builds in telemetry becomes a priority.

PR #562, opened 2026-07-01
feat(csharp): default SEA result compression to LZ4 (derived from cloudfetch.lz4.enabled)

eric-wang-1990:
## Summary

Make the ADBC C# **Statement Execution (SEA/REST)** path request **LZ4 result compression by default**, matching the driver's own Thrift/CloudFetch path and the JDBC / databricks-sql-python connectors. Instead of a hardcoded default, the SEA `result_compression` default is **derived from the pre-existing LZ4 capability flag** `adbc.databricks.cloudfetch.lz4.enabled` (`CanDecompressLz4`, default `true`), so a single flag drives LZ4 across both the Thrift and REST paths and disabling it keeps disabling LZ4 everywhere.

Previously `StatementExecutionConnection` only set `result_compression` when `adbc.databricks.rest.result_compression` was explicitly provided; otherwise the field was omitted (`null` → serialized-away) and the server returned **uncompressed** Arrow. This was an outlier:

| Driver / path | Default SEA `result_compression` |
|---|---|
| **ADBC C# SEA (before this PR)** | omitted (uncompressed) |
| ADBC C# Thrift / CloudFetch | LZ4 (`CanDecompressLz4 = true`) |
| JDBC (`databricks-jdbc`) | `LZ4_FRAME` for query/metadata; **`NONE` for updates** (see below) |
| `databricks-sql-python` | `LZ4_FRAME` (`enable_query_result_lz4_compression=True`) |

## Change

- `StatementExecutionConnection`: default `_resultCompression` to `CanDecompressLz4 ? "LZ4_FRAME" : "NONE"` — derived from `adbc.databricks.cloudfetch.lz4.enabled` (default `true`).
- Precedence: an explicit `adbc.databricks.rest.result_compression` always wins over the derived default (set to `NONE` to opt out).
- Compression is only meaningful for `ARROW_STREAM`; for `JSON_ARRAY` / `CSV` the default is `NONE` even when the capability flag is enabled (an explicit override still wins).
- Updated the parameter doc comment and `csharp/README.md` to reflect the new default and accurate accepted values (`LZ4_FRAME` / `NONE`).

## ExecuteUpdate fix (LZ4-framed `num_affected_rows`)

Requesting LZ4 by default surfaced a latent bug in the SEA update path. `ExecuteUpdate` read the inline `num_affected_rows` result with a **hand-rolled parser** that fed `result.Attachment` straight into `ArrowStreamReader` **without checking the manifest for LZ4**. Once LZ4 was requested, the server returned that 1-row result LZ4-framed; the raw parse threw, a bare `catch` swallowed it, and the affected-row count came back as `-1` — failing the `ExecuteUpdate_{Insert,Update,Delete}Data_ReturnsAffectedRows` E2E tests.

Fix: route `ExecuteUpdate` through the shared **`CreateReader`** path (the same one the query path uses), which already centralizes LZ4 decompression, inline-vs-external-links, and multi-chunk decoding, then read `num_affected_rows` from the first record batch. This also fixes the latent case where a DML result arrives via external links (the old attachment-only parser silently returned `-1`).

## Why ADBC decompresses rather than opting out of compression for updates

`databricks-jdbc` avoids compressing update results entirely: on its SEA path, `UPDATE` is excluded from CloudFetch, so it uses `format=JSON_ARRAY` + `disposition=INLINE` and forces `result_compression=NONE` (`DatabricksSdkClient.getRequest`). **ADBC cannot mirror this** — the ADBC reader only handles **Arrow**, so it always requests `ARROW_STREAM` and cannot fall back to `JSON_ARRAY` for updates.

Given that constraint, the two options were:
- **(A) [chosen]** request LZ4 uniformly and decompress the update result through the shared `CreateReader`;
- (B) special-case the update path to force `INLINE` + `NONE`.

We chose **A** because it fixes the general class of bug (one Arrow decode path, driven by the response manifest, for query/update/metadata) rather than reintroducing per-statement-type request divergence. It also matches how JDBC itself behaves on its *hybrid* (`INLINE_OR_EXTERNAL_LINKS`) query path, where it likewise requests LZ4 and decompresses inline results based on the server's reported compression. The cost of A (compressing an ~8-byte `num_affected_rows` result once per DML statement) is negligible.

The read path continues to select decompression from the **response manifest** (`response.Manifest?.ResultCompression == "LZ4_FRAME"`), so responses the server chooses to return uncompressed still work.

## Tests

Unit tests in `StatementExecutionResultCompressionTests` (mock at the `HttpMessageHandler` level):
- `ExecuteStatement_DefaultsToLz4FrameCompression` — no property set, capability flag defaults true → `result_compression = LZ4_FRAME`.
- `ExecuteStatement_RespectsExplicitCompressionOverride` — explicit `NONE` → `NONE`.
- `ExecuteStatement_DerivesNoneWhenLz4Disabled` — `cloudfetch.lz4.enabled=false`, no override → `NONE`.
- `ExecuteStatement_ExplicitCompressionOverridesLz4DisabledFlag` — explicit `LZ4_FRAME` with flag disabled → `LZ4_FRAME` (override wins).
- `ExecuteStatement_DerivesNoneForNonArrowFormat` (`JSON_ARRAY`, `CSV`) → `NONE`.
- `ExecuteStatement_ExplicitCompressionHonoredForNonArrowFormat` → explicit override still wins for non-Arrow format.
- `ExecuteUpdate_ReadsLz4FramedNumAffectedRows` — SUCCEEDED response with an LZ4-framed inline `num_affected_rows` result → `ExecuteUpdate` returns the real count (reproduces the CI E2E failure at unit level).

Local run: `Passed: 8, Failed: 0` (net8.0).


jadewang-db (approved):

csharp/src/StatementExecution/StatementExecutionConnection.cs
```diff
@@ -334,7 +334,21 @@ private void ValidateProperties()
...
+            // flag keeps disabling LZ4 everywhere. An explicit adbc.databricks.rest.result_compression
+            // takes precedence over the derived default. The reader decompresses based on the
+            // response manifest, so an uncompressed response is still handled correctly.
+            //
+            // Result compression in the SEA API is only meaningful for ARROW_STREAM; for
+            // JSON_ARRAY/CSV the field was previously omitted, so only derive the LZ4 default
+            // when the format is ARROW_STREAM. This preserves the prior (working) behavior for
+            // users who explicitly select a non-Arrow format without setting a compression.
+            bool canDecompressLz4 = PropertyHelper.GetBooleanPropertyWithValidation(properties, DatabricksParameters.CanDecompressLz4, true);
+            bool isArrowStream = _resultFormat.Equals("ARROW_STREAM", StringComparison.OrdinalIgnoreCase);
```
eric-wang-1990: You do not need to add this isArrowStream check, we should just set _resultCompression directly, it will be passed to backend.

csharp/src/StatementExecution/StatementExecutionConnection.cs
```diff
@@ -334,7 +334,21 @@ private void ValidateProperties()
...
+            string defaultResultCompression = (canDecompressLz4 && isArrowStream) ? "LZ4_FRAME" : "NONE";
+            _resultCompression = PropertyHelper.GetStringProperty(properties, DatabricksParameters.ResultCompression, defaultResultCompression);
```
eric-wang-1990: Should we just deprecate ResultCompression? No client is using that and it seem cause unnecessary confusion.

eric-wang-1990: Yes let's deprecate/remove result_compression

csharp/src/StatementExecution/StatementExecutionConnection.cs
```diff
@@ -334,7 +334,24 @@ private void ValidateProperties()
...
+            // no real client and only added confusion. The LZ4 capability flag is the single knob.
+            //
+            // LZ4_FRAME is only requested for Arrow-based result formats. The SEA server's
+            // tolerance of LZ4_FRAME on non-Arrow formats (JSON_ARRAY, CSV) is not something we
+            // verify here, and this driver's reader is Arrow-only so non-Arrow results are not
+            // consumable regardless. Gating on format keeps the request conservative: non-Arrow
+            // formats send NONE (matching the prior behavior) rather than relying on the server
+            // to silently ignore a compression codec it may not accept for that format.
+            bool canDecompressLz4 = PropertyHelper.GetBooleanPropertyWithValidation(properties, DatabricksParameters.CanDecompressLz4, true);
+            bool isArrowFormat = _resultFormat.IndexOf("ARROW", StringComparison.OrdinalIgnoreCase) >= 0;
```
eric-wang-1990: Let's remove the isArrowFormat check, we can always send LZ4 compression, if it is not arrow then it is just ignored on server side

PR #563, opened 2026-07-01
fix(csharp): apply cached feature flags on connect (warm cache)

jadewang-db:
PR #500 changed the connection-open feature-flag merge to fire-and-forget: `MergeWithEnvironmentConfigAndFeatureFlags` kicked off the async merge in a discarded `Task.Run` and returned the env-config-only properties, so server feature flags (e.g. `adbc.databricks.enable_fast_metadata_query`) were never applied to any connection. The synchronous applier existed but had no callers, so neither the first nor any subsequent connection ever received the flags.

Add `FeatureFlagCache.TryMergeWarmFeatureFlags`: if the per-host cache is warm, merge the cached flags synchronously (local properties win); if cold, kick off a background warm-up and return null so `Connect()` proceeds unblocked and a subsequent connection to the same host picks up the flags. Never performs a blocking network fetch, and honors the negative-cache backoff.

Co-authored-by: Isaac

## What's Changed

Closes #NNN.


csharp/src/FeatureFlagCache.cs
```diff
@@ -412,6 +412,56 @@ public IReadOnlyDictionary<string, string> MergePropertiesWithFeatureFlags(
...
+        /// <returns>The merged properties when the cache is warm; otherwise <c>null</c>.</returns>
+        public IReadOnlyDictionary<string, string>? TryMergeWarmFeatureFlags(
+            IReadOnlyDictionary<string, string> localProperties,
+            string assemblyVersion)
+        {
+            // Respect the enable flag (default true; only an explicit, parseable "false" disables).
+            if (localProperties.TryGetValue(DatabricksParameters.FeatureFlagCacheEnabled, out string? enabledStr) &&
+                bool.TryParse(enabledStr, out bool parsedEnabled) && !parsedEnabled)
+            {
+                return null;
```
eric-wang-1990: Can we return existing localProperties for all these null returns? Since this is public method TryMergeWarmFeatureFlags and we should always return some valid properties?

eric-wang-1990 (approved):

csharp/src/FeatureFlagValueValidator.cs
```diff
@@ -0,0 +1,126 @@
...
+            PositiveLong,
+        }
+
+        // Strictly-typed connection parameters, keyed by their property name (== the flag name the
+        // service returns). Mirrors the PropertyHelper.Get*PropertyWithValidation call sites.
+        private static readonly IReadOnlyDictionary<string, ValueKind> s_typedParameters =
+            new Dictionary<string, ValueKind>
+            {
+                // Boolean (GetBooleanPropertyWithValidation)
+                [DatabricksParameters.ApplySSPWithQueries] = ValueKind.Boolean,
```
eric-wang-1990: This mean we need to maintain this list when new proerties are coming in.
The better way for this is to move the validator to the property definition, so when a property is added the validator is registered with it. This can be a follow up though

PR #566, opened 2026-07-02
fix(csharp): derive SEA wait_timeout from direct-results, treat CLOSED as success (ES-2034600)

eric-wang-1990:
## Summary

The SEA / Statement Execution path sent `wait_timeout="10s"` by default (whenever direct results were enabled). A **non-zero `wait_timeout` combined with `INLINE_OR_EXTERNAL_LINKS`** routes the request to the old SEA sync-hybrid results path, which **truncates multi-chunk results**: the manifest advertises N chunks, but only chunk 0 is delivered (no `external_links`, no next-chunk pointers) and `GET /result/chunks/{n>=1}` returns `400 "chunk_index n is out of bounds"`. The client silently receives a subset of rows — e.g. `GetColumns` returning **0 rows**. Root cause tracked server-side as **ES-2034600**.

## Change (mirrors `databricks-jdbc`)

Derive `wait_timeout` from the direct-results flag instead of exposing it as a customer knob:
- **direct results ON (default)** → `wait_timeout` **unset** → server returns the full result inline in one response (`SUCCEEDED` or `CLOSED`) — the SEA equivalent of Thrift DirectResults, and what `databricks-jdbc` sends by default.
- **direct results OFF** → `wait_timeout="0s"` (fully async) → poll.
- **Never** send a positive `wait_timeout`. `adbc.databricks.rest.wait_timeout` is now ignored (kept only as a recognized-but-inert key).

Also:
- **Treat `state=CLOSED` as success** — read the direct result from that response and mark the statement closed so `Dispose` skips a redundant `CloseStatement`. Matches `databricks-jdbc`, which treats `CLOSED == SUCCEEDED`.
- **Decouple the metadata-operation timeout** (`CreateMetadataTimeoutCts`) from `wait_timeout` with a fixed 300s bound — fixes a latent `FromSeconds(0)` near-immediate cancel when direct results were disabled.

This is independent of (and more fundamental than) the SEA LZ4-default work in #562 — it makes the driver safe regardless of compression.

## Tests

- **Unit** (`StatementExecutionWaitTimeoutTests`): direct-results-on → `wait_timeout` omitted; direct-results-off → `"0s"`; `adbc.databricks.rest.wait_timeout` ignored; `CLOSED` response with a result yields rows (not an exception). Full unit suite **875/875**.
- **E2E** (live SEA warehouse): `GetColumns` false-case returned **148,282 rows** via the direct-result `CLOSED` path (previously truncated under sync+LZ4); `StatementExecutionDriverE2ETests` **12/12** (queries/updates/DDL).
- Curl-verified: unset `wait_timeout` → `CLOSED` + full inline result (1 chunk); non-zero sync → 2-chunk truncation.

### DIFF
--- a/csharp/src/StatementExecution/StatementExecutionConnection.cs
+++ b/csharp/src/StatementExecution/StatementExecutionConnection.cs
@@ -59,7 +59,9 @@ namespace AdbcDrivers.Databricks.StatementExecution
         private string _resultDisposition = null!;
         private string _resultFormat = null!;
         private string? _resultCompression;
-        private int _waitTimeoutSeconds;
+        // Request wait_timeout: null = unset (direct results on), "0s" = async (direct results off).
+        // Never a positive value (see ValidateProperties / ES-2034600).
+        private string? _waitTimeout;
         private int _pollingIntervalMs;
         private bool _enablePKFK;
         private bool _enableMultipleCatalogSupport;
@@ -332,12 +334,18 @@ namespace AdbcDrivers.Databricks.StatementExecution
             _resultFormat = PropertyHelper.GetStringProperty(properties, DatabricksParameters.ResultFormat, "ARROW_STREAM");
             properties.TryGetValue(DatabricksParameters.ResultCompression, out _resultCompression);
 
-            _waitTimeoutSeconds = PropertyHelper.GetIntPropertyWithValidation(properties, DatabricksParameters.WaitTimeout, 10);
-            if (properties.TryGetValue(DatabricksParameters.EnableDirectResults, out var directResults) &&
-                directResults.Equals("false", StringComparison.OrdinalIgnoreCase))
-            {
-                _waitTimeoutSeconds = 0;
-            }
+            // wait_timeout is NOT a customer-tunable knob; it is derived from the direct-results flag.
+            //   Direct results ON (default): leave wait_timeout UNSET so the server returns the full
+            //     result inline in a single response (state SUCCEEDED or CLOSED) — the SEA equivalent
+            //     of Thrift DirectResults, and what databricks-jdbc sends by default.
+            //   Direct results OFF: send wait_timeout="0s" (fully async) and poll for the result.
+            // We deliberately never send a NON-ZERO wait_timeout: a positive value routes the request
+            //   to the old SEA sync-hybrid results path, which truncates multi-chunk results (the
+            //   manifest advertises N chunks but only chunk 0 is delivered). Tracked as ES-2034600.
+            bool enableDirectResults =
+                !(properties.TryGetValue(DatabricksParameters.EnableDirectResults, out var directResults)
+                  && directResults.Equals("false", StringComparison.OrdinalIgnoreCase));
+            _waitTimeout = enableDirectResults ? null : "0s";
             // PECO-3064: adbc.apache.statement.polltime_ms is the single key for SEA polling cadence
             // (consolidated with the Thrift path). SEA defaults to 1000 ms — HTTP/JSON polls are
             // heavier than Thrift's 100 ms default — but both protocols share the same property.
@@ -547,7 +555,7 @@ namespace AdbcDrivers.Databricks.StatementExecution
                 _resultDisposition,
                 _resultFormat,
                 _resultCompression,
-                _waitTimeoutSeconds,
+                _waitTimeout,
                 _pollingIntervalMs,
                 _properties,
                 _recyclableMemoryStreamManager,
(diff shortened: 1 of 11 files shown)


csharp/src/StatementExecution/StatementExecutionConnection.cs
```diff
@@ -983,9 +991,14 @@ internal async Task<List<RecordBatch>> ExecuteShowColumnsAsync(
             return _catalog;
         }
 
+        // Metadata operations (GetObjects) can legitimately take a while (full-catalog SHOW COLUMNS,
+        // etc.), so bound them with a generous fixed timeout — decoupled from the request wait_timeout
+        // (which is now unset/"0s"; FromSeconds(0) would have cancelled metadata almost immediately).
+        private const int MetadataOperationTimeoutSeconds = 300;
```
eric-wang-1990: Why are we hardcoding 300s here, what are other metadata operaton timeout like passed into IGetObjectsDataProvider.GetCatalogsAsync?

csharp/src/StatementExecution/StatementExecutionConnection.cs
```diff
@@ -983,9 +1002,16 @@ internal async Task<List<RecordBatch>> ExecuteShowColumnsAsync(
             return _catalog;
         }
 
+        // Metadata operations (GetObjects/GetTableSchema) are just queries — bound them with the same
+        // query timeout as regular queries (adbc.apache.statement.query_timeout_s, default 3h,
+        // matching Thrift). A single metadata call can fan out into many SHOW COLUMNS / SHOW CATALOGS
+        // statements, so this bounds the whole tree. 0 = no timeout (never-cancelled CTS), matching
+        // PollWithTimeoutAsync and the Thrift ApacheUtility.GetCancellationToken semantics.
         internal CancellationTokenSource CreateMetadataTimeoutCts()
```
eric-wang-1990, line 1032: I think we do not need this timeout. Since for SEA/REST all the metadata queries is going through ExecuteQueryAsyncInternal, which is just like execute a query directly. And for that path the query_timeout is already applied.

csharp/src/StatementExecution/StatementExecutionStatement.cs
```diff
@@ -369,9 +378,13 @@ private async Task<QueryResult> ExecuteQueryInternalAsync(CancellationToken canc
             {
                 throw new AdbcException("Statement execution was canceled");
             }
+            // CLOSED = execution succeeded and the statement is already closed server-side; the
+            // direct result (manifest + inline attachment) is present in THIS response. Treat it
+            // like SUCCEEDED — read the result below — and remember it's closed so Dispose skips the
+            // redundant CloseStatement. Matches databricks-jdbc, which treats CLOSED == SUCCEEDED.
```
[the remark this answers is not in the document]
eric-wang-1990: This is the public doc for returne status:
[status](https://docs.databricks.com/api/gcp/workspace/statementexecution/getstatement#status)
object
The status response includes execution state and if relevant, error information.
[error](https://docs.databricks.com/api/gcp/workspace/statementexecution/getstatement#status-error)
object
[error_code](https://docs.databricks.com/api/gcp/workspace/statementexecution/getstatement#status-error-error_code)
string
Enum: UNKNOWN | INTERNAL_ERROR | TEMPORARILY_UNAVAILABLE | IO_ERROR | BAD_REQUEST | SERVICE_UNDER_MAINTENANCE | WORKSPACE_TEMPORARILY_UNAVAILABLE | DEADLINE_EXCEEDED | CANCELLED | RESOURCE_EXHAUSTED | ABORTED | NOT_FOUND | ALREADY_EXISTS | UNAUTHENTICATED
[message](https://docs.databricks.com/api/gcp/workspace/statementexecution/getstatement#status-error-message)
string
A brief summary of the error condition.
[sql_state](https://docs.databricks.com/api/gcp/workspace/statementexecution/getstatement#status-sql_state)
string
SQLSTATE error code returned when the statement execution fails. Only populated when the statement status is FAILED.
[state](https://docs.databricks.com/api/gcp/workspace/statementexecution/getstatement#status-state)
string
Enum: PENDING | RUNNING | SUCCEEDED | FAILED | CANCELED | CLOSED
Statement execution state:
PENDING: waiting for warehouse
RUNNING: running
SUCCEEDED: execution was successful, result data available for fetch
FAILED: execution failed; reason for failure described in accompanying error message
CANCELED: user canceled; can come from explicit cancel call, or timeout with on_wait_timeout=CANCEL
CLOSED: execution successful, and statement closed; result no longer available for fetch

csharp/src/StatementExecution/StatementExecutionConnection.cs
eric-wang-1990: That is valid, what does Thrift path set as metadata timeout in this case? We should be consistent

eric-wang-1990: No Thrift GetObjects is also many thrift calls, you need to see the submodule of hiveServer2 to see how it works.

eric-wang-1990: I added the hiveserver2 as a context repo, please try again, you should be able to access it now

csharp/src/StatementExecution/StatementExecutionStatement.cs
```diff
@@ -177,11 +181,12 @@ public StatementExecutionStatement(
             _resultDisposition = resultDisposition ?? throw new ArgumentNullException(nameof(resultDisposition));
             _resultFormat = resultFormat ?? throw new ArgumentNullException(nameof(resultFormat));
             _resultCompression = resultCompression;
-            _waitTimeoutSeconds = waitTimeoutSeconds;
+            _waitTimeout = waitTimeout;
             _pollingIntervalMs = pollingIntervalMs;
             _properties = properties ?? throw new ArgumentNullException(nameof(properties));
+            // Default matches the Thrift path (3h) so SEA queries aren't unbounded; 0 = no timeout.
```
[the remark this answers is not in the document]
eric-wang-1990: It is intended, maybe update the PR description to reflect that

jadewang-db (approved):

Issue #525, opened 2026-06-16
metadata(csharp): _/% wildcard semantics and TABLE_CAT identifier echo diverge between Thrift and SEA


eric-wang-1990:
Part of #519 (Thrift vs SEA metadata parity).

## Problem

The `_` (single-char) and `%` SQL-LIKE wildcards are interpreted differently by the two protocols, and the returned catalog/schema identifiers are echoed differently.

**Identifier echo** (~1,700 diffs): querying catalog `comparator_tests` (underscore):

```
TABLE_CAT / TABLE_CATALOG: thrift="comparator-tests" → sea="comparator_tests"
   list_columns 1,078x · list_tables 600x · list_schemas 13x
```

Thrift treats `_` as a single-char wildcard, so `comparator_tests` matches the hyphenated catalog `comparator-tests` and Thrift returns the **server's matched name**; SEA treats `_` literally and **echoes the caller's argument**. This also drives a large share of the "row only on side1/side2" presence diffs (whole rows appear on one side because the two protocols match different object sets).

`catalog="%"` on `list_schemas` similarly makes Thrift enumerate all catalogs (31 extra schema rows) while SEA does not expand the same way.

## Expected

A single, documented wildcard semantics for `_`/`%`, applied identically on both paths, and a single rule for whether `TABLE_CAT`/`TABLE_SCHEM` echoes the caller's argument or the server's matched name.

## How JDBC does it

JDBC converts patterns via `WildcardUtil.jdbcPatternToHive` (`_`→`.`, `%`→`*`, `\_`→literal) on the SEA path; the Thrift path passes through to HiveServer2. (Note: JDBC is itself **not** fully consistent here — this issue should establish the single intended ADBC behavior rather than copy JDBC verbatim.) For echo, JDBC echoes the caller's catalog argument for `TABLE_CAT` while returning the server's actual schema/table names.

## Suggested fix

Define and centralize (in #519) the wildcard escaping + identifier-echo rules so both protocols match the same objects and return the same `TABLE_CAT`/`TABLE_SCHEM`.

## Evidence

`diff-report-20260615-064733Z.json`, `where = .../ TABLE_CAT | TABLE_CATALOG` and `rows[*] (only on side1/side2)`.

Issue #519, opened 2026-06-16
Thrift vs SEA metadata parity: comparator surfaced ~8.3K diffs across metadata ops (tracking)


eric-wang-1990:
## Summary

The Thrift-vs-SEA comparator harness (`databricks-driver-test/tests/comparator-tests/csharp-adbc`) compared the two protocols of the **same** C# ADBC Databricks driver across the JDBC-derived metadata + statement matrices. The latest run (`results-20260615-064733Z`) produced **8,346 diffing test cases / 35,853 individual cell-or-schema differences** out of ~12.8K cases.

The diffs are **not random** — they collapse into ~10 systematic behavioral divergences between the Thrift metadata path and the SEA/REST metadata path, amplified across the large cartesian matrices (`list_columns` = 6,561 cases, `list_tables` = 5,103). Every single `list_columns` case diffs, driven by a handful of fixed schema/field differences.

## Root cause (architectural)

The Databricks **JDBC** driver routes *both* its Thrift response and its SEA/SQL-EXEC response through **one shared normalization layer** — `MetadataResultSetBuilder` (`dbclient/impl/common/MetadataResultSetBuilder.java`) plus the column schema in `common/MetadataResultConstants.java`. That layer:

- defines the JDBC column schema once (names, Arrow/JDBC types, non-null map),
- **derives** `SQL_DATA_TYPE`, `CHAR_OCTET_LENGTH`, `BUFFER_LENGTH`, `COLUMN_SIZE`, `SQL_DATETIME_SUB` from the type name via a single `getCode()` / size lookup,
- normalizes `ORDINAL_POSITION` to 1-based, and `NULLABLE`/`IS_NULLABLE` from one source field,
- runs identically whether the raw rows came from Thrift (`getThriftRows`) or SEA (`getRows`).

The ADBC C# driver has **no equivalent normalization layer** — each protocol path emits whatever the server/Thrift RPC returns, so the two diverge. The strategic fix is to introduce an ADBC analog: a shared metadata-shaping component both protocol readers feed into, so Thrift and SEA produce byte-identical JDBC-shaped metadata.

## Findings → tracked sub-issues

- [ ] #520 — `list_columns` `BUFFER_LENGTH` Arrow type `int8` (Thrift) vs `int32` (SEA) — **5,541 diffs**
[...]

PR #574, opened 2026-07-08
fix(csharp): treat catalog as a literal name on the Thrift metadata path (#525)

eric-wang-1990:
## Summary

Makes the **Thrift** metadata path treat the `catalog` argument of `GetSchemas`/`GetTables`/`GetColumns` as an **exact-name identifier**, not a SQL-LIKE pattern — matching the SEA path and PR #536.

Today the catalog is sent to the server's Thrift `GetSchemas`/`GetTables`/`GetColumns` RPC as a pattern, so `_`/`%` act as wildcards and **over-match**. Per the JDBC/ODBC contract (catalog is an exact name; only schema/table/column are patterns) and for Thrift↔SEA parity, catalog should match literally.

## Change

`DatabricksStatement` normalizes the catalog after SPARK handling and before the base RPC call (`NormalizeCatalogForThrift`):

- **Bare match-all wildcard** (`%` or `*`, like null) → `null` ("all catalogs"), mirroring the SEA path (`StatementExecutionStatement.EffectiveCatalog`) / #536.
- **Any other catalog name** → escape `_`/`%` so the server matches it literally. The Thrift server honors the `\_`/`\%` escape.
- When `escape_pattern_wildcards` is already enabled, the base `HiveServer2Statement` escapes catalog itself — so we return it unescaped here to **avoid double-escaping**.

The `AdbcDrivers.HiveServer2` submodule is **not** modified; the change lives entirely in the Databricks-owned `DatabricksStatement` overrides.

## Evidence (live Thrift warehouse)

Verified against a live Thrift SQL warehouse via `GetSchemas`:

| catalog arg | matched | 
|---|---|
| `None` / `%` / `*` | all catalogs |
| `comparator_tests` (unescaped `_`) | **comparator-tests AND comparator_tests** (over-match) |
| `comparator\_tests` (escaped) | **comparator_tests only** (literal) |

So escaping produces the correct literal match, and `%`/`*` still return all catalogs.

## Test plan

- [x] Unit: `DatabricksStatementUnitTests.NormalizeCatalogForThrift_NormalizesAndEscapesCatalog` — 11 cases: `%`/`*`/null→null, plain name unchanged, `_`/`%` escaped (flag off), no double-escape (flag on).
- [ ] CI runs unit + E2E (the change targets net10.0 via the submodule; couldn't build locally on SDK 9.0 — will validate in CI).

## Notes

- Part of #519 (Thrift↔SEA metadata parity); companion to #536 (SEA `%`/`*`→null). Suggest merging after #536 so the pair reads together.
- Minor behavior change: multi-catalog **disabled** + `catalog="%"` now returns the session default catalog's schemas (was empty), intentionally matching SEA/#536.

### DIFF
--- a/csharp/src/DatabricksStatement.cs
+++ b/csharp/src/DatabricksStatement.cs
@@ -672,6 +672,35 @@ namespace AdbcDrivers.Databricks
         }
 
         /// <summary>
+        /// Normalizes the catalog argument for the Thrift metadata path (GetSchemas/GetTables/GetColumns)
+        /// so that catalog behaves as an exact-name identifier rather than a SQL-LIKE pattern (issue #525),
+        /// matching the SEA path and PR #536.
+        ///
+        /// - A bare match-all wildcard ("%" or "*") — like a null argument — means "all catalogs", so it is
+        ///   mapped to null. On Thrift the server already expands "%"/"*", but mapping to null keeps the
+        ///   behavior identical to the SEA path (StatementExecutionStatement.EffectiveCatalog).
+        /// - Any other catalog name has its "_" and "%" wildcard characters escaped so the Thrift server
+        ///   matches it literally. The server honors the "\_"/"\%" escape (verified against a live warehouse);
+        ///   without escaping it treats "_" as a single-char wildcard, so e.g. catalog "comparator_tests"
+        ///   would also match "comparator-tests".
+        ///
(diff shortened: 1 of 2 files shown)


eric-wang-1990:
Superseded by #536 (fix */% wildcard semantics and TABLE_CAT identifier echo between Thrift and SEA) — same issue #525 this PR targeted, landed there for both protocols. Reopen if a Thrift catalog-literal gap remains.

PR #575, opened 2026-07-08
perf(csharp): stream CloudFetch LZ4 decompression to bound multi-chunk partial-read memory

### LINKED ISSUE #573: SEA receives UNCOMPRESSED external CloudFetch results under INLINE_OR_EXTERNAL_LINKS (more transfer + memory than Thrift)
## Confirmed root cause (SEA-specific)

Live-warehouse diagnostic on `SELECT * FROM main.tpcds_sf1_delta.store_sales` (2.88M rows, 65 external chunks), logging the CloudFetch downloader's runtime `isLz4Compressed`:

- **SEA (rest), driver default `INLINE_OR_EXTERNAL_LINKS`:** `isLz4=False` — the server returns external result files **uncompressed** (~21 MB raw chunks). `manifest.ResultCompression != "LZ4_FRAME"` at runtime.
- **Thrift:** `isLz4=True` — external chunks are LZ4-compressed (~5 MB each).

(Note: an explicit `EXTERNAL_LINKS`-disposition probe *does* return `LZ4_FRAME`, so the compression appears tied to the disposition the driver requests.)

**Consequence:** for the same result, SEA downloads and holds **uncompressed** external data, while Thrift transfers compressed. Partial-read peak-held memory (read one batch + dispose, prefetch active): SEA ~98 MB vs Thrift's compressed footprint. SEA does more network transfer and holds more memory.

## Relationship to #575

#575 (stream CloudFetch LZ4 decompression) reduces peak memory for the *compressed* multi-chunk path — measured **Thrift 103 → 26 MB (−75%)**. It does **not** help SEA today because SEA's external files are uncompressed (nothing to decode-lazily). Once SEA receives LZ4-compressed external results, it benefits from #575 automatically.

## What to investigate here

Why SEA doesn't get LZ4-compressed external results under `INLINE_OR_EXTERNAL_LINKS` despite requesting `result_compression=LZ4_FRAME`, and how to fix it (e.g. disposition choice for large results, or how compression is requested for the hybrid/external path). Goal: SEA external CloudFetch results come back LZ4-compressed like Thrift — cutting transfer volume and (with #575) partial-read memory.

## History / corrections

The original framing (inline reader eager materialization) and a later "SEA retains 3× Thrift" claim were both measurement artifacts (`GC.GetTotalMemory` noise; single-chunk `RANGE`). The deterministic re-measurement (`GetTotalAllocatedBytes` + peak-held on the 65-chunk `store_sales`) is the reliable basis for the above.

This issue was written by Isaac.

eric-wang-1990:
## What

`CloudFetchDownloader` decompressed each downloaded chunk **in full up front** (`Lz4Utilities.DecompressLz4Async` → a whole `RecyclableMemoryStream`) before the reader consumed anything. With CloudFetch prefetch downloading several chunks ahead, a **partial read** (read one batch, then dispose — e.g. connection-pool reuse) held *every prefetched chunk fully decompressed* in memory.

This decodes **lazily**: new `Lz4Utilities.CreateDecompressingStream(byte[], ArrayPool)` wraps the compressed bytes in the existing forward-only `CustomLZ4DecoderStream`, and `CloudFetchDownloader` hands that to the reader. Prefetched-but-unread chunks are held **compressed** and only decoded as the reader reads them; decompressed bytes are released as consumed. Size/telemetry accounting is retargeted to the compressed size (decompressed size isn't known up front), and the now-unused `RecyclableMemoryStreamManager` field is removed.

## Measured improvement (live warehouse)

`SELECT * FROM main.tpcds_sf1_delta.store_sales` (2.88M rows, 23 cols, **65 external chunks**), partial read = one batch then dispose. Peak managed memory held while the reader is alive:

| Protocol | before (eager) | after (streaming) |
|---|---|---|
| **Thrift** | 103.2 MB | **25.7 MB** (−75%) |
| **SEA (rest), on #562** | 67.7 MB | **17.4 MB** (−74%) |

Both protocols see ~75% lower peak memory on multi-chunk partial reads. Full read returns all rows unchanged (decoding intact); the path is shared and stays correct for both.

## Dependency on #562

The SEA number is measured on the **#562 branch**, because SEA only requests LZ4 (`result_compression=LZ4_FRAME`) once #562 lands. On plain `main` (pre-#562) SEA receives external results **uncompressed** (`isLz4=False`), so there's nothing to decode-lazily and this change is a no-op for SEA. Once #562 merges, SEA benefits automatically — verified above. (Thrift already requests LZ4, so it benefits immediately.) Ideally land after #562.

## Testing

- Full read of `store_sales` returns all rows on both protocols; existing SEA/Thrift E2E + StatementExecution unit suites pass; builds clean under `TreatWarningsAsErrors`.
- Before/after peak-held measured via forced-GC sampling while the reader is alive; deterministic-allocation cross-checks agree.

Fixes #573

### DIFF
--- a/csharp/src/Lz4Utilities.cs
+++ b/csharp/src/Lz4Utilities.cs
@@ -102,6 +102,32 @@ namespace AdbcDrivers.Databricks
         }
 
         /// <summary>
+        /// Wraps LZ4-framed compressed bytes in a forward-only stream that decompresses
+        /// <b>incrementally as it is read</b>, rather than decompressing the whole payload up front.
+        /// </summary>
+        /// <remarks>
+        /// Unlike <see cref="DecompressLz4(byte[], RecyclableMemoryStreamManager, ArrayPool{byte})"/>,
+        /// which fully materializes the decompressed result before returning, this hands back a lazy
+        /// decoder over the compressed bytes. When the consumer (e.g. <c>ArrowStreamReader</c>) reads
+        /// only part of the result and disposes, only the batches actually read are decompressed —
+        /// the whole result is never materialized. Peak/allocation is bounded to what is consumed
+        /// plus the (smaller) compressed buffer. The caller MUST dispose the returned stream; disposal
+        /// closes the inner compressed-bytes stream. See #573.
+        /// </remarks>
+        /// <param name="compressedData">The LZ4-framed compressed bytes.</param>
+        /// <param name="bufferPool">The ArrayPool for the decoder's work buffers (from DatabricksDatabase).</param>
+        /// <returns>A forward-only, self-decompressing <see cref="Stream"/>. Caller must dispose.</returns>
+        public static Stream CreateDecompressingStream(byte[] compressedData, ArrayPool<byte> bufferPool)
+        {
+            return new CustomLZ4DecoderStream(
+                new MemoryStream(compressedData),
+                descriptor => descriptor.CreateDecoder(),
+                bufferPool,
+                leaveOpen: false,
+                interactive: false);
+        }
+
+        /// <summary>
         /// Asynchronously decompresses LZ4 compressed data into memory.
         /// Returns a RecyclableMemoryStream that must be disposed by the caller.
         /// </summary>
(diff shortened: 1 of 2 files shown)


eric-wang-1990:
This will cause partial result, closing

PR #609, opened 2026-08-02
feat(csharp): validate exact-match args for SEA GetPrimaryKeys / GetCrossReference

eric-wang-1990:
## Summary

Adds client-side argument validation to the SEA metadata **exact-match** operations
`GetPrimaryKeys` and `GetCrossReference`, mirroring the JDBC reference driver's
`resolveKeyBasedParams` / `listCrossReferences`. This gives clean, deterministic
errors and avoids relying on the Thrift server's inconsistent (and in one case buggy)
handling of missing arguments.

Split out of adbc-drivers/databricks#604 (which handles object-not-found → empty,
backslash escaping, and the exception-type pivot) because exact-match argument
validation is a distinct concern.

## Verified behavior (live thrift vs rest, + JDBC reference source)

Legend: **empty** = 0-row result, no throw. **throw** = AdbcException.

### GetPrimaryKeys(catalog, schema, table)

| input | Thrift (measured) | JDBC SEA (resolveKeyBasedParams) | ADBC SEA (this PR) |
|---|---|---|---|
| all set, object exists | rows | rows | rows |
| all set, object missing | **empty** | **empty** (isObjectNotFound catch) | **empty** |
| **table = null** | throw `42000` "tableName may not be null" | **throw** "tableName may not be null" | **throw `42000`** |
| **catalog set, schema = null** | throw `08000` "GET_FUNCTIONS assertion failed" (server bug) | **throw** "schema may not be null when catalog is specified" | **throw `42000`** (clean; matches JDBC intent) |
| catalog = null (schema/table set) | throw `08000` (server bug) | resolves from session, else throw | empty (no session default) |

### GetCrossReference(pk…, fk…)

| input | Thrift (measured) | JDBC SEA (listCrossReferences) | ADBC SEA (this PR) |
|---|---|---|---|
| foreign side fully set, exists | rows | rows | rows |
| **both parent table AND foreign table = null** | **throw `42000`** "foreignTable and parentTableName are both null" | **empty** ("unspecified") | **empty** (matches JDBC SEA — see note) |
| **≥1 table set, foreign table = null** (e.g. parent table set) | **empty** | **empty** ("unspecified") | **empty** |
| **foreign catalog set, foreign schema = null** (fk table set) | throw `08000` "GET_FUNCTIONS assertion failed" (server bug) | **throw** "schema may not be null when catalog is specified" | **throw `42000`** (clean; matches JDBC intent) |
| PK/FK feature disabled | (n/a) | empty | **empty** (short-circuit before validation; symmetric with GetPrimaryKeys) |
| object missing | empty | empty (isObjectNotFound catch) | empty |

> **Note on the both-null case:** ADBC SEA follows the **JDBC SEA** reference here, not Thrift.
> `DatabricksMetadataQueryClient.listCrossReferences` checks *only* the foreign table — a null
> foreign table is "unspecified" and returns an empty result; it never inspects the parent table.
> So both-tables-null returns **empty**. Measured against the live Thrift server, the same input
> throws a clean `42000` ("foreignTable and parentTableName are both null"), so this is an
> intentional Thrift-vs-SEA divergence, whitelisted in the comparator (databricks-driver-test#1047).
> When only the foreign table is null but a parent table is set, both Thrift and SEA return empty.

## Why validate client-side

1. **Thrift's `08000` on null-schema is a server bug** — it surfaces an internal
   `GET_FUNCTIONS assertion failed` (a connection-exception SQLSTATE) instead of a
   clean argument error. JDBC validates before the RPC and throws a clean error;
   this PR does the same so SEA doesn't depend on (or replicate) the buggy path.
2. **Deterministic, no round-trip** — the error is raised before issuing SHOW KEYS /
   SHOW FOREIGN KEYS.

## Scope / correctness notes

- Validation lives only on the **user-facing** command path. Each exact-match op is a
  thin public wrapper that reads the statement fields, validates, and throws
  (`GetPrimaryKeysAsync` / `GetCrossReferenceAsync`), over a **non-throwing** internal
  core that takes the key parts as explicit params and returns empty for
  unspecified/invalid args (`GetPrimaryKeysAsyncInternal(catalog, schema, table, …)` /
  `GetCrossReferenceAsyncInternal(pkCatalog, pkSchema, pkTable, fkCatalog, fkSchema, fkTable, …)`).
  `GetColumnsExtended`'s internal three-call fallback reuses the PK/FK fetch with a null
  schema on purpose (to gather keys for a column set); it calls the **internals**
  directly, so its legitimately-unspecified args return empty instead of being rejected.
  The wrappers throw only when the feature is engaged and the internal would otherwise
  reach a live fetch, so the two paths never diverge.
- Errors are `DatabricksException` (the SEA-natural type). The comparator compares
  `Status` + `SqlState`, not the concrete subclass, so this matches Thrift's path
  where Thrift returns a clean error, and is whitelisted where Thrift emits its
  `08000` bug.

## Comparator

The residual Thrift `08000`-vs-SEA `42000` sqlstate diffs (foreign/catalog schema-null)
are documented as a Thrift server bug in databricks-driver-test's test_case_filters
(schema-null exact-match skip). The both-tables-null case is also whitelisted there
(databricks-driver-test#1047): ADBC SEA now returns **empty** (matching JDBC SEA) while
Thrift throws `42000` — an intentional Thrift-vs-SEA divergence.

All behavior in the tables above was verified against the live Thrift server via a
throwaway E2E probe (`TGetPrimaryKeysReq` / `TGetCrossReferenceReq`), not inferred from
code or mocks.

### DIFF
--- a/csharp/src/StatementExecution/StatementExecutionStatement.cs
+++ b/csharp/src/StatementExecution/StatementExecutionStatement.cs
@@ -1660,10 +1660,18 @@ namespace AdbcDrivers.Databricks.StatementExecution
         // so its legitimately-unspecified args return empty rather than being rejected.
         private async Task<QueryResult> GetPrimaryKeysAsync(CancellationToken cancellationToken)
         {
-            // Validate only when the PK/FK feature is actually engaged — the internal short-circuit
-            // (ShouldReturnEmptyPKFKResult) otherwise returns empty for these args, and the public
-            // command must match that (don't throw for a request the feature would no-op anyway).
-            if (!MetadataUtilities.ShouldReturnEmptyPKFKResult(_metadataCatalogName, null, _connection.EnablePKFK))
+            // Throw only in the case the internal would otherwise reach a live fetch for: the
+            // feature engaged (else the internal short-circuits to empty via ShouldReturnEmptyPKFKResult),
+            // AND one of the two exact-match args the JDBC reference driver rejects client-side is bad
+            // (a missing table, or a catalog-set-schema-null request). These preconditions mirror the
+            // throws inside the block so an ordinary successful call — valid catalog with table and
+            // schema both set — never enters the block and never emits a spurious empty span.
+            bool willReject =
+                string.IsNullOrEmpty(_metadataTableName)
+                || (!string.IsNullOrEmpty(_metadataCatalogName) && string.IsNullOrEmpty(_metadataSchemaName));
+
+            if (!MetadataUtilities.ShouldReturnEmptyPKFKResult(_metadataCatalogName, null, _connection.EnablePKFK)
+                && willReject)
             {
                 // Emit a span for a client-side rejection so the failure is captured in telemetry —
                 // the throw happens before GetPrimaryKeysAsyncInternal, so its own activity (with
--- a/csharp/src/StatementExecution/StatementExecutionStatement.cs
+++ b/csharp/src/StatementExecution/StatementExecutionStatement.cs
@@ -1580,22 +1580,11 @@ namespace AdbcDrivers.Databricks.StatementExecution
             if (columnsResult.Stream == null)
                 return columnsResult;
 
-            // This is the internal columns-extended reuse, not the user-facing
-            // getprimarykeys command: a null schema (or other exact-match args the
-            // user-facing command rejects) is legitimate here. GetPrimaryKeysAsync
-            // throws on those, so guard and fall back to an empty PK result rather
-            // than letting the validation throw. (For a null schema GetPrimaryKeysAsync
-            // would return an empty result anyway, so this is behavior-preserving.)
-            QueryResult pkResult;
-            if (string.IsNullOrEmpty(_metadataTableName)
-                || (!string.IsNullOrEmpty(_metadataCatalogName) && string.IsNullOrEmpty(_metadataSchemaName)))
-            {
-                pkResult = MetadataSchemaFactory.CreateEmptyPrimaryKeysResult();
-            }
-            else
-            {
-                pkResult = await GetPrimaryKeysAsync(cancellationToken).ConfigureAwait(false);
-            }
+            // Internal columns-extended reuse, not the user-facing getprimarykeys command:
+            // a null schema (legitimately passed here while gathering PKs for a column set)
+            // must NOT be rejected. Call the non-throwing internal, which returns an empty PK
+            // result for such "unspecified" args instead of throwing.
+            var pkResult = await GetPrimaryKeysAsyncInternal(cancellationToken).ConfigureAwait(false);
 
             // Find FKs where the current table is the FK (child) side — null PK params to
             // match any parent, mirroring Thrift's GetCrossReferenceAsForeignTableAsync.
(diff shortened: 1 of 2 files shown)


csharp/src/StatementExecution/StatementExecutionStatement.cs
```diff
@@ -1639,6 +1660,22 @@ private async Task<QueryResult> GetPrimaryKeysAsync(CancellationToken cancellati
...
                     return MetadataSchemaFactory.CreateEmptyPrimaryKeysResult();
 
+                // GetPrimaryKeys is an exact-match operation. Validate required args
+                // client-side (mirroring the JDBC reference driver's resolveKeyBasedParams):
+                //   - table null/empty          -> throw "tableName may not be null"
+                //   - catalog set + schema null -> throw "schema may not be null when catalog is specified"
+                // Validating here (before issuing SHOW KEYS) also avoids the Thrift server's
+                // internal "GET_FUNCTIONS assertion failed" bug on schema-null, and gives a
+                // clean, deterministic error instead of relying on a server round-trip.
+                if (validateArgs)
```
eric-wang-1990: Instead of passing in this extra validateArgs param, we should change the callsite where we do not expect it to throw, to return empty rather than throw.

csharp/src/StatementExecution/StatementExecutionStatement.cs
```diff
@@ -1564,7 +1580,22 @@ private async Task<QueryResult> GetColumnsExtendedViaThreeCalls(CancellationToke
...
 
-            var pkResult = await GetPrimaryKeysAsync(cancellationToken).ConfigureAwait(false);
+            // This is the internal columns-extended reuse, not the user-facing
+            // getprimarykeys command: a null schema (or other exact-match args the
+            // user-facing command rejects) is legitimate here. GetPrimaryKeysAsync
+            // throws on those, so guard and fall back to an empty PK result rather
+            // than letting the validation throw. (For a null schema GetPrimaryKeysAsync
+            // would return an empty result anyway, so this is behavior-preserving.)
+            QueryResult pkResult;
+            if (string.IsNullOrEmpty(_metadataTableName)
```
eric-wang-1990: Should this be changed to try catch pattern? So that if there are other throw scenario in the future we do not break this return empty thing?

eric-wang-1990: Refactored (2e5fd7d) per the suggestion: instead of the caller-side pre-check guard (or the earlier `validateArgs` flag), each exact-match op is now a thin **public** wrapper that validates + throws, delegating to a **non-throwing internal** core:

- `GetPrimaryKeysAsync` / `GetCrossReferenceAsync` (public commands) validate the exact-match args and throw a clean 42000, then call the internal.
- `GetPrimaryKeysAsyncInternal` / `GetCrossReferenceAsyncInternal` do the fetch and return **empty** for unspecified/invalid args (never throw). `GetColumnsExtendedViaThreeCalls` calls these directly, so its legitimately-null schema is never rejected.

This avoids the try/catch (which would have to catch `DatabricksException`/`InternalError`/42000 broadly and could swallow a real server error), and it's future-proof: any new validation added to a public wrapper stays out of the internal reuse path by construction. Verified behavior-identical to the guard version across all catalog/schema/table combinations; 240 StatementExecution unit tests pass.

csharp/src/StatementExecution/StatementExecutionStatement.cs
```diff
@@ -1627,23 +1650,61 @@ await ProcessRelationshipDataSafe(
...
         }
 
+        // The user-facing getprimarykeys command. GetPrimaryKeys is an exact-match operation,
+        // so this wrapper validates the required args client-side (mirroring the JDBC reference
+        // driver's resolveKeyBasedParams) and throws a clean 42000 on a missing table / a
+        // catalog-set-schema-null request — instead of relying on the Thrift server round-trip
+        // (which surfaces an internal "GET_FUNCTIONS assertion failed" 08000 bug on schema-null).
+        // The actual fetch lives in GetPrimaryKeysAsyncInternal, which does NOT throw; the
+        // internal columns-extended reuse (GetColumnsExtendedViaThreeCalls) calls that directly
+        // so its legitimately-unspecified args return empty rather than being rejected.
```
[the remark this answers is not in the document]
eric-wang-1990: Follow-up (027779d): kept this telemetry fix but tidied the form — swapped the async lambda + filler `await Task.CompletedTask` for the synchronous `TraceActivity(Action<Activity?>)` overload (it already sets Error status, records the exception event, and rethrows). Same span (`GetPrimaryKeys`/`GetCrossReference`) with the same tags on rejection; identical exception contract; 240 StatementExecution tests pass.

csharp/src/StatementExecution/StatementExecutionStatement.cs
```diff
@@ -1627,23 +1650,81 @@ await ProcessRelationshipDataSafe(
...
+            // feature engaged (else the internal short-circuits to empty via ShouldReturnEmptyPKFKResult),
+            // AND one of the two exact-match args the JDBC reference driver rejects client-side is bad
+            // (a missing table, or a catalog-set-schema-null request). These preconditions mirror the
+            // throws inside the block so an ordinary successful call — valid catalog with table and
+            // schema both set — never enters the block and never emits a spurious empty span.
+            bool willReject =
+                string.IsNullOrEmpty(_metadataTableName)
+                || (!string.IsNullOrEmpty(_metadataCatalogName) && string.IsNullOrEmpty(_metadataSchemaName));
+
+            if (!MetadataUtilities.ShouldReturnEmptyPKFKResult(_metadataCatalogName, null, _connection.EnablePKFK)
```
eric-wang-1990: MetadataUtilities.ShouldReturnEmptyPKFKResult should be the first check to return empty. Then you can also skip the same check inside GetPrimaryKeysAsyncInternal

PR #611, opened 2026-08-03
TEMP diag: PartialRead pool retention probe (do not merge)

eric-wang-1990:
Temporary diagnostic to prove whether the PartialRead ~40MB growth on the merge-queue runner is RecyclableMemoryStreamManager pool retention (LZ4) vs a real leak. Dumps pool Large/Small Free/InUse sizes at before/after snapshots, LZ4 on vs off. Not for merge.


eric-wang-1990:
Closing — diagnostic yielded no usable data (OutputHelper output is suppressed by the xUnit console logger on passing tests, and the probe had no assertion). Not needed anyway: the test fix in #612 subtracts the RMSM free-pool bytes directly, so it's robust to the pool peak regardless of magnitude.

PR #612, opened 2026-08-03
test(csharp): make PartialRead leak check pool-aware (exclude RMSM free list)

eric-wang-1990:
## Summary

Fixes the intermittent merge-queue failure of `DriverResilienceTests.PartialRead_DisposeStatement_ShouldNotHangOrLeak` (~40 MB "growth" vs its 20 MB threshold), which has flaked on real E2E runs since #562 made SEA request LZ4 by default.

**It is not a leak.** LZ4 decompression composes its output stream from pooled 128 KB blocks (`RecyclableMemoryStreamManager`); on dispose those blocks return to the manager's **free list**, which `GC.GetTotalMemory` counts as live. That free list is (a) legitimately reusable memory and (b) timing-dependent on a partial read — how many CloudFetch chunks decompressed before the pipeline was cancelled. So a raw GC delta is inherently flaky. The old single full-read warm-up under-sized the pool relative to the 5-iteration measurement loop, so the pool's own growth landed inside the measurement window and read as a leak.

## Fix (test-only)

- **Subtract the manager's free-pool bytes** (`SmallPoolFreeSize + LargePoolFreeSize`) from GC total at both snapshots, so the assertion measures memory that is neither in use nor parked in the reusable pool — i.e. a genuine leak. Immune to how large the reuse pool grows under load.
- **Warm up with the same partial-read cycle** the measurement loop uses (6×), so the pool reaches steady state before the baseline snapshot.

Threshold (20 MB) and intent (catch a real per-iteration leak of abandoned CloudFetch buffers) are unchanged. No driver code changes.

## Verification

- Local (SEA): non-pooled growth **0.04 MB** (pool free 7.88 MB correctly excluded).
- Validated on a real merge-queue-class E2E run via the `e2e-test` label (see checks).

Note: the *product* question of bounding the `RecyclableMemoryStreamManager` free list (its default is unbounded, and the manager is per-`DatabricksDatabase` and never disposed — relevant for long-lived multi-tenant hosts like PowerBI) is tracked separately; this PR only corrects the test's measurement.

### DIFF
--- a/csharp/test/E2E/DriverResilienceTests.cs
+++ b/csharp/test/E2E/DriverResilienceTests.cs
@@ -67,19 +67,31 @@ namespace AdbcDrivers.Databricks.Tests
             // total at both snapshots: what remains is memory that is neither live-in-use nor
             // parked in the reusable pool — i.e. genuinely leaked. This is immune to how large the
             // reuse pool happens to grow under load.
+            // DIAGNOSTIC MODE (temporary — attributes the ~24MB/5-iter non-pooled growth to a
+            // specific pool). Uses Console.WriteLine (NOT OutputHelper — the xUnit console logger
+            // suppresses OutputHelper output for PASSING tests, which is why the earlier probe
+            // produced no data). Per snapshot reports:
+            //   GC  = GC.GetTotalMemory(true)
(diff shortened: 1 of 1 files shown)


eric-wang-1990:
Closing — the pool-free-subtraction approach was insufficient (the growth is LOH heap high-water-mark, not the RMSM free list). Superseded by #614, which uses a WeakReference invariant.

PR #613, opened 2026-08-03
TEMP diag: isolate PartialRead in E2E (do not merge)

eric-wang-1990:
Temporary: injects TEST_FILTER to run only PartialRead_DisposeStatement_ShouldNotHangOrLeak in the E2E job. If it passes isolated but fails in the full suite, the growth is process/pool carryover from earlier tests, not this test's own leak. Not for merge.


eric-wang-1990:
Closing — diagnostic branch (TEST_FILTER isolation + object trackers). Its job is done: it proved the failure is not a leak (WeakReference aliveReaders=0/7; GC.GetTotalMemory plateau is LOH heap accounting). Real fix is #614.

PR #620, opened 2026-08-04
fix(csharp): SEA PK/XREF reject only null schema, not empty-string (match JDBC SEA)

eric-wang-1990:
## Summary

SEA PK/FK exact-match validation now rejects only a **null** schema (matching the JDBC reference driver), not an empty-string schema.

## The bug

`GetPrimaryKeysAsync` / `GetCrossReferenceAsync` used `string.IsNullOrEmpty(schema)` to throw the client-side `"schema may not be null when catalog is specified"` (SQLSTATE `42000`). This over-validates versus JDBC SEA: `DatabricksMetadataQueryClient.resolveKeyBasedParams` rejects only `schema == null`. An **empty-string** schema is a different case — JDBC passes it through to the server, which rejects it with object-not-found (`42704`), and the driver maps that to an **empty result** (`isObjectNotFoundException`).

So for `catalog=main, schema="", table=t`:
- **JDBC SEA / server / ADBC Thrift**: server responds `42704` → empty (JDBC) / throws `42704` (Thrift).
- **ADBC SEA (before)**: threw a generic client-side `42000` before ever reaching the server — the outlier.

## The fix

Change the schema guard from `IsNullOrEmpty(schema)` to `schema == null` in both `GetPrimaryKeysAsync` and `GetCrossReferenceAsync`. Now:
- **null schema + catalog set** → still throws clean `42000` (unchanged, matches JDBC).
- **empty-string schema** → passes through to the server → `42704` → the non-throwing internal's `IsObjectNotFoundException` catch (which already treats `42704` as not-found) returns **empty** — matching JDBC SEA.

The **table** guard keeps `IsNullOrEmpty` (JDBC rejects null-*or-empty* table — `table == null || table.isEmpty()`).

## Impact on the comparator

Closes the `get_primary_keys` / `get_cross_reference` **sqlstate** divergence (server/Thrift `42704` vs SEA client-side `42000`) for empty-string-schema inputs — ADBC SEA now returns empty like JDBC SEA. (The residual Thrift-throws-vs-SEA-empty *outcome* divergence for these inputs is whitelisted in databricks-driver-test as the same JDBC-aligned pattern as the other object-not-found cases.)

## Test plan

- [x] Added regression tests: `GetPrimaryKeys_CatalogSetSchemaEmptyString_ReturnsEmpty` and `GetCrossReference_ForeignCatalogSetForeignSchemaEmptyString_ReturnsEmpty` (empty-string schema → empty, via a mocked `42704`).
- [x] Existing null-schema throw tests (`GetPrimaryKeys_CatalogSetSchemaNull_Throws`, etc.) still pass — null still throws `42000`.
- [x] 243 StatementExecution unit tests pass.
- [ ] Merge-queue E2E + comparator run.

### DIFF
--- a/csharp/src/StatementExecution/StatementExecutionStatement.cs
+++ b/csharp/src/StatementExecution/StatementExecutionStatement.cs
@@ -1674,11 +1674,11 @@ namespace AdbcDrivers.Databricks.StatementExecution
 
                 // Match JDBC SEA (resolveKeyBasedParams): the schema check rejects only a
                 // *null* schema when a catalog is specified — NOT an empty-string schema. An
-                // empty-string schema is passed through to the server, which rejects it with a
-                // clean object-not-found (SQLSTATE 42704); the internal's IsObjectNotFoundException
-                // catch then maps that to an empty result (as JDBC SEA does). Using IsNullOrEmpty
-                // here would over-validate — throwing a generic 42000 client-side for empty schema
-                // where JDBC (and thus the SEA contract) returns empty.
+                // empty-string schema falls through to GetPrimaryKeysAsyncNoThrow, whose
+                // IsNullOrEmpty(catalog/schema/table) short-circuit returns an empty result
+                // client-side (no SHOW KEYS issued) — matching JDBC SEA's empty result. Using
+                // IsNullOrEmpty here would over-validate — throwing a generic 42000 client-side
+                // for empty schema where JDBC (and thus the SEA contract) returns empty.
                 if (!string.IsNullOrEmpty(_metadataCatalogName) && _metadataSchemaName == null)
                     throw NewInvalidArgumentException("schema may not be null when catalog is specified");
             }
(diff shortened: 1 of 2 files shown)


eric-wang-1990:
Closing — superseded by the comparator whitelist (databricks-driver-test#1089), consistent with the decision to resolve the empty-string-identifier metadata divergences at the comparator level rather than by touching the driver again.

Context: this PR relaxed the SEA PK/XREF schema guard from `IsNullOrEmpty(schema)` to `schema == null` so an empty-string schema returns empty (matching JDBC SEA's `resolveKeyBasedParams`, which checks only `schema == null`) instead of throwing a client-side 42000. That's a faithful JDBC-SEA correction, but:

1. **The comparator gap is already closed by #1089**, which whitelists the empty-string schema divergence (`{schema: ""}` / `{foreign_schema: ""}`) as an intentional Thrift-vs-SEA difference. This PR changes no comparator outcome — even with it, ADBC SEA (empty) still differs from ADBC Thrift (throws 42704) on empty-string schema, so the whitelist entry stands either way.
2. It re-touches the exact-match validation that just landed in #609 for an uncommon edge (empty-string, as opposed to null, schema).

Leaving the empty-string over-validation as a documented minor known-gap. If we later want the driver to faithfully match JDBC SEA here, the one-line `== null` change in this PR is the fix to revive.

